PR #6, opened 2021-11-23
EHPR-25 | Lint/Package setup

arranfw:
image


packages/web/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  plugins: ['jsx-a11y', 'prettier'],
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+    'plugin:react/recommended',
+    'plugin:react-hooks/recommended',
+    'plugin:jsx-a11y/recommended',
+    '../../.eslintrc.js',
```
hitankar-freshworks, line 9: do we need to reference the parent eslint config?

hitankar-freshworks: I think if it make sense to even have eslint in root.

arranfw: My thinking is that the root config is useful to keep consistency in the codebases, if we decided we wanted a rule disabled in the web app I think it would make sense to also follow that rule in the API, a root config allows us to set rules for each package while also letting packages use specific plugins

fwpushan (approved):
LGTM 👍

arranfw:
Let's get a couple more eyes on this and then add husky in another PR, this one's big enough 😄

packages/web/.eslintrc.js
hitankar-freshworks: okay. makes sense.

hitankar-freshworks (approved):

packages/api/package.json
```diff
@@ -37,17 +37,17 @@
...
     "@types/express": "^4.17.13",
-    "@types/jest": "27.0.2",
+    "@types/jest": "26.0.14",
     "@types/node": "^16.0.0",
     "@types/supertest": "^2.0.11",
     "eslint": "^8.0.1",
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^4.0.0",
-    "jest": "^27.2.5",
+    "jest": "26.6.0",
```
dbayly-freshworks, line 46: Nit but don't we want to be using the latest version of jest

packages/api/package.json
```diff
@@ -37,17 +37,17 @@
     "@nestjs/schematics": "^8.0.0",
     "@nestjs/testing": "^8.0.0",
     "@types/express": "^4.17.13",
```
dbayly-freshworks, line 39: Nit, but maybe we should take the time to lockdown all these packages

dbayly-freshworks (approved):
Looks good save for a pair of nits

packages/api/package.json
arranfw: I was getting package resolution errors with CRA/React scripts when the versions were mismatched. There was an option to skip preflight checks and ignore the warnings which I admittedly don't fully understand the consequences of.

I valued consistency in versions across packages over having the latest version, what do you think?

packages/api/package.json
arranfw: Probably smart, thanks!

vyasworks (approved):
LGTM

vyasworks (commented):
can we use es2020

tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
-  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "module": "commonjs",
     "declaration": true,
```
vyasworks, line 4: can we use es2020

packages/api/package.json
dbayly-freshworks: It should be fine, since this is a dev dependency. This does add some potential tech debt if we want to update our packages in the future.

tsconfig.json
arranfw: Sure, do you have a resource or a quick explanation on your preference?

package.json
```diff
@@ -7,6 +7,7 @@
     "format:write": "yarn prettier --write .",
     "lint": "echo 'Linting all packages' && wsrun --parallel --exclude-missing lint",
     "test": "echo 'Running tests' && wsrun --parallel --exclude-missing test",
+    "build": "echo 'Building all packages' && wsrun --parallel --exclude-missing build",
```
fwpushan, line 10: Don't run build in parallel ... run in `--stages`, it will pickup package with no internal dependencies. Otherwise dependent package will break

fwpushan (changes requested):
Please check my comments.

package.json
arranfw: 🧠 totally!

tsconfig.json
fwpushan: @vyasworks 
1. Do Node 16.* support es2020?
2. Do Lambda support Node 16?

fwpushan (approved):

tsconfig.json
arranfw: https://stackoverflow.com/a/67371788

Looks like es2020 won't be a problem for node 16

arranfw: A brief look [here](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) shows node  14 for lambdas which looks like it supports es2020 

https://stackoverflow.com/a/61305579

vyasworks: Yes supported natively. That is a good SO thread too.

vyasworks: we can use node14 - https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases

this is recommended config

tsconfig.json
```diff
@@ -1,20 +1,23 @@
...
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es2017",
     "sourceMap": true,
     "incremental": true,
     "skipLibCheck": true,
+    "strict": true,
     "strictNullChecks": false,
     "noImplicitAny": false,
```
vyasworks, line 16: Should be `true`

arranfw: isn't it :o

arranfw: `strictNullChecks` and `noImplicitAny` could be true

tsconfig.json
arranfw: `"extends": "@tsconfig/node12/tsconfig.json",`
should we add that?

tsconfig.json
```diff
@@ -1,20 +1,23 @@
...
     "removeComments": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
-    "target": "es2017",
     "sourceMap": true,
     "incremental": true,
     "skipLibCheck": true,
+    "strict": true,
     "strictNullChecks": false,
```
vyasworks, line 15: strictNullChecks could be true, since this is brand new project.

tsconfig.json
vyasworks: No, sorry I meant - https://www.npmjs.com/package/@tsconfig/node14

There is a link to Node 14 in the above URL pls check that

arranfw: I should have waited to merge 🤕

PR #8, opened 2021-11-25
EHPR-46 | Husky + Lint

arranfw:
https://freshworks.atlassian.net/browse/EHPR-46

Currently opting to rely on editor tooling and CI steps to check lint rather than using `lint-staged`, it's causing some errors that I can't find much information on. Happy to re-add it if we can figure it out

For now just prettier with husky


dbayly-freshworks (approved):
Looks ok to me

vyasworks (approved):

hitankar-freshworks (approved):

fwpushan (commented):
Nice job, I will suggest add lint support for common package or execute lint from root of repo.

PR #9, opened 2021-11-25
Ehpr 5 mass email send

dbayly-freshworks:
This is still very much a WIP . I welcome your feedback! 

Current features include: 
* Authentication and sending emails with Ches 
* generating mock csv with 10k rows and uuids 
* asynchronous sending 
* rate limiting
* Automatic refresh and retry when a token expires
* Store failed results in a csv 
* Mocked out random delay in sending


fwpushan (changes requested):
My suggestions  are
1. Put scripts as different package. Api is a node application with an entry-point. Also each scripts are standalone node applications.
2.  If there is any reusable functionality, put it either in common package or separate package.

.gitignore
```diff
@@ -12,3 +12,4 @@ dist
 node_modules
 .DS_Store
 .env
+**/*.env
```
fwpushan, line 15: Put all env variables in root. Because env will be injected from root of the repo.

packages/api/src/scripts/emailSend.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+  try {
+    const response = await axios.post(CHES_AUTH_URL, params, config);
+    return response.data.access_token;
+  } catch (error) {
+    console.error(error);
+    return error;
+  }
+}
+async function updateToken() {
```
arranfw, line 39: Could this be renamed to `getChesToken` or something to be more accurate for what it does?

packages/api/src/scripts/generatecsv.js
```diff
@@ -0,0 +1,23 @@
+const fs = require('fs');
+const uuidV4 = require('uuid').v4;
+const faker = require('faker');
```
arranfw, line 3: We need to install these packges somewhere right?

arranfw (commented):
Curious about the decision to write this in JS

The script in HCAP had retry functionality for network disruptions or other intermittent failures, could we add that here? It could possibly reduce failed sends. Or do we just intend to re-run the script on the list of failed sends afterwards?

fwpushan (changes requested):
As per new project structure. This is not reusable component. Please put it into apps.

Otherwise LGTM.

```suggestion
This is standalone local script. We can get input as cmd argument (instead of ENV) for that we can use this https://www.npmjs.com/package/minimist
```

dbayly-freshworks:
Closing this as it needs a rework. Will reopen when it is ready.

PR #10, opened 2021-11-26
Added tests to workflow

dbayly-freshworks:


arranfw:
GH Actions expert right here :P What do you think of running the build in PRs as well? LTC had some more complex workflows that would build a package if there is a diff in it. Could help catch errors outside of linting/testing

arranfw (approved):

fwpushan (approved):

dbayly-freshworks:
> GH Actions expert right here :P What do you think of running the build in PRs as well? LTC had some more complex workflows that would build a package if there is a diff in it. Could help catch errors outside of linting/testing

Won't the projects get built as part of the tests? I'm all for it in principle though.

PR #13, opened 2021-11-26
EHPR-50 | Add basic service and script to start local application stack

fwpushan:
Add basic support to start local application


.github/workflows/pr-check.yml
```diff
@@ -24,3 +24,5 @@ jobs:
         run: yarn run check
       - name: Run all tests
         run: yarn run test
+      - name: Build All packages
+        run: make build
```
arranfw, line 28: Wondering if we can split these up so changes in /api don't trigger a build for /web

Makefile
```diff
@@ -7,17 +7,18 @@ export $(shell sed 's/=.*//' ./.env)
 export NODE_ENV ?= development
 
 
-.PHONY: build-local api-local web-local print-env run-local
+.PHONY: build api-local web-local print-env run-local
```
arranfw, line 10: What's this .PHONY thing?

Makefile
```diff
@@ -26,8 +27,18 @@ api-local:
...
 web-local:
-	@echo "++\n***** Running Web in local Node server\n ++"
+	@echo "++\n***** Running Web in local Node server\n++"
 	@yarn workspace @ehpr/web start
-	@echo "\n*****"
+	@echo "++\n*****"
+
+start-local-services:
+	@echo "++\n***** Starting local services\n++"
+	@docker-compose up db
```
arranfw, line 36: Why just db? And can we add --build to this command

dbayly-freshworks (approved):
Looks fine to me

Makefile
```diff
@@ -7,17 +7,18 @@ export $(shell sed 's/=.*//' ./.env)
 export NODE_ENV ?= development
 
 
-.PHONY: build-local api-local web-local print-env run-local
+.PHONY: build api-local web-local print-env run-local
 
-run-local: | build-local api-local web-local
+run-local: | build api-local web-local
```
arranfw, line 12: Do we need to build before running local?

Makefile
fwpushan: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Makefile
fwpushan: here we are running underlying services specified in docker-compose not building anything.

Makefile
fwpushan: here build step ensure all build artefacts are upto date.  This is just building each package locally.

arranfw: I think we can save considerable time by only building /common here

Makefile
arranfw: Cool, over my head 😄

Makefile
fwpushan: there could be more than one common packages. How to ensure build them before building/running api and web?

arranfw: Possibly another script? `build-dependencies` that only builds required packages

vyasworks: we should look into `watch` options that can look at files changing and run `tsc` for us, this way, frequent changes on any packages can be used in the other.

arranfw (approved):

Makefile
fwpushan: Added watch for common folder and keep watch mode for api and web as built-in script for packages.

package.json
```diff
@@ -11,7 +11,11 @@
     "build": "echo 'Building all packages' && wsrun --stages --exclude-missing build",
     "check": "run-p 'lint' 'format:check'",
     "prepare": "husky install",
-    "precommit": "run-p 'format:staged'"
+    "precommit": "run-p 'format:staged'",
+    "watch:commons": "nodemon --exec \"yarn build && yarn\"",
```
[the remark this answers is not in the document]
arranfw: Do you know if there is a performance difference between using nodemon to build vs using `tsc --watch`?

packages/web/package.json
```diff
@@ -4,10 +4,10 @@
   "private": true,
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build-app": "react-scripts build",
```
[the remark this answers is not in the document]
fwpushan: > I think we can save considerable time by only building /common here
I try to address this comment. api + web both have respective watch mode. So, building locally is not required. We have to think about some naming of build script for common packages.

package.json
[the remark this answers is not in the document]
fwpushan: almost, same then we have to add separate watch script for each packages.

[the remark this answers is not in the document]
fwpushan: and need to run yarn to link modified packages with dependent packages.

.github/workflows/pr-check.yml
fwpushan: DONE

package.json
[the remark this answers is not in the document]
fwpushan: Fixed naming

package.json
```diff
@@ -11,7 +11,11 @@
...
+    "start:api:local": "yarn workspace @ehpr/api start:dev:local",
+    "start:web:local": "yarn workspace @ehpr/web start",
```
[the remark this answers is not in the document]
fwpushan: already rebased

Makefile
fwpushan: tsc --watch is working fine, so change this to tsc --watch

Makefile
```diff
@@ -7,9 +7,7 @@ export $(shell sed 's/=.*//' ./.env)
 export NODE_ENV ?= development
 
 
-.PHONY: build-common watch-common app-local print-env run-local start-local-services
-
-run-local: | build-common app-local 
+.PHONY: build-common watch-common app-local print-env start-local-services
```
arranfw, line 10: Still not sure if we need to build-common before watch-common

arranfw: Can we make these more consistent?
```suggestion
.PHONY: print-env watch-common start-local-apps start-local-services
```

package.json
```diff
@@ -8,11 +8,11 @@
     "format:write": "yarn prettier --write .",
     "lint": "echo 'Linting all packages' && wsrun --parallel --exclude-missing lint",
     "test": "echo 'Running tests' && wsrun --parallel --exclude-missing test",
-    "build": "echo 'Building all packages' && wsrun --stages --exclude-missing build",
+    "build-common": "echo 'Building all packages' && wsrun --stages --exclude-missing build-pre-app",
```
arranfw, line 11: What is build-pre-app? Looks like a dupe of build

package.json
```diff
@@ -8,11 +8,11 @@
...
     "check": "run-p 'lint' 'format:check'",
     "prepare": "husky install",
     "precommit": "run-p 'format:staged'",
-    "watch:common": "nodemon --exec \"yarn build && yarn\"",
+    "watch:common": "wsrun --stages --exclude-missing watch",
```
arranfw, line 15: This looks like it would spawn 1 watch process per package, I think what @vyasworks was suggesting is a single watch at the root level

arranfw: @vyasworks Do we need to `watch` api/web? Or just common packages

arranfw: Nest/Next provide their own `watch`es right?

vyasworks: Do we want watch for each package or just a common one? It would be easier and better if we had the ability to get local running with minimum commands. 

I am imagining something like this: 

# - DockerCompose Run - `Backing services` OR `Backing Services + API + Web` 
# - Start command in each module?
# - Yarn watch in the root

fwpushan: @vyasworks `nodemon` was creating single watch for common packages. But that was not supper performance oriented .....

fwpushan: @vyasworks and I will work on Dockerfile for creation of whole application execution with watch.

packages/api/package.json
```diff
@@ -28,6 +28,7 @@
     "@nestjs/common": "8.0.0",
     "@nestjs/core": "8.0.0",
     "@nestjs/platform-express": "8.0.0",
+    "dotenv": "^10.0.0",
```
dbayly-freshworks, line 31: Nit, might be good to lock down this package here

package.json
fwpushan: this is to decoupling common packages build from all builds ....

arranfw: Why two names for the same thing? `common`/`pre-app`

package.json
```diff
@@ -8,7 +8,7 @@
     "format:write": "yarn prettier --write .",
     "lint": "echo 'Linting all packages' && wsrun --parallel --exclude-missing lint",
     "test": "echo 'Running tests' && wsrun --parallel --exclude-missing test",
-    "build-common": "echo 'Building all common packages packages' && wsrun --stages --exclude-missing build-pre-app",
+    "build-common": "echo 'Building all common packages packages' && wsrun --stages --exclude @ehpr/api --exclude @ehpr/web --exclude-missing build",
```
arranfw, line 11: Last comment, we can fix this later but I'd like the watch script to follow this pattern as well, being able to explicitly say "watch all packages except API/Web"

arranfw (approved):
Good stuff

PR #14, opened 2021-11-26
EHPR-6 | Add tailwind

arranfw:


dbayly-freshworks (approved):
This may be radical, but I say merge it and deal with any issues that pop up later :sonic:

arranfw:
> This may be radical, but I say merge it and deal with any issues that pop up later :sonic:

Totally agree

PR #17, opened 2021-11-27
EHPR-6 | Stepper

arranfw:
image

image

### DIFF
--- a/packages/web/pages/submission/[step].tsx
+++ b/packages/web/pages/submission/[step].tsx
@@ -1,0 +1,40 @@
+import Link from 'next/link';
+import { useRouter } from 'next/router';
+
+import { Form, Stepper } from '@components';
+
+const FORM_STEPS = ['Primary', 'Contact', 'Credential', 'Preferences', 'Review and Submit'];
+
+const Submission = () => {
+  const router = useRouter();
+
+  const step = Number(router.query.step);
+
+  return (
+    <Form>
+      <div className='w-full'>
+        <Stepper formSteps={FORM_STEPS} step={step} />
+      </div>
+      <div className='flex justify-center'>
+        <Link href={step === 1 ? '/submission/1' : `/submission/${Number(step) - 1}`}>
+          <a type='button' className='w-full md:w-40 px-8 py-2 btn btn-outline md:mb-0 mr-0 '>
+            Go Back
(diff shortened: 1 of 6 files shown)


packages/web/pages/submission/[step].tsx
```diff
@@ -0,0 +1,46 @@
...
+            className='w-full md:w-40 px-8 py-2 btn btn-outline md:mb-0 mr-0 text-center'
+          >
+            Go Back
+          </a>
+        </Link>
+        <Link
+          href={
+            step === FORM_STEPS.length
+              ? `/submission/${FORM_STEPS.length}`
+              : `/submission/${Number(step) + 1}`
```
Comment (dbayly-freshworks) [inline packages/web/pages/submission/[step].tsx:31]: This should work for now, but I wonder if there's  a more elegant way of handling the routing.

dbayly-freshworks (approved):
One small comment, but otherwise looks great!

packages/web/pages/submission/[step].tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const Submission = () => {
+  const router = useRouter();
+
+  const step = Number(router.query.step);
+
+  return (
+    <Form>
+      <div className='w-full p-4 border-b'>
+        <Stepper formSteps={FORM_STEPS} step={step} />
```
Comment (fwpushan) [inline packages/web/pages/submission/[step].tsx:16]: Stepper is not interactive ?

Comment (arranfw) [reply packages/web/pages/submission/[step].tsx:16]: I think that was the outcome of our discussions, to keep it like Imms form for simplicity for now

packages/web/pages/submission/[step].tsx
Comment (fwpushan) [reply packages/web/pages/submission/[step].tsx:31]: You can programatically invoke history.* method for routing.

fwpushan (approved):
LGTM, only few minor comments.

packages/web/pages/submission/[step].tsx
Comment (fwpushan) [reply packages/web/pages/submission/[step].tsx:16]: OK

packages/web/pages/submission/[step].tsx
Comment (arranfw) [reply packages/web/pages/submission/[step].tsx:31]: Happy to abstract this to a function, I was considering changing this to a button instead of a link as it's more of a functional action rather than a navigation action. I'll do that

PR #18, opened 2021-11-27
Created data model for form

dbayly-freshworks:
This is my take on an interface for all the form data. I welcome your feedback.

### DIFF
--- a/packages/api/constants/form-structure.ts
+++ b/packages/api/constants/form-structure.ts
@@ -1,0 +1,83 @@
+export interface FormData {
+  // Page 1
+  first_name: string;
+  last_name: string;
+  postal_code: string;
+  // Page 2
+  primary_phone: string;
+  secondary_phone: string;
+  email: string;
+  // Page 3
+  stream_types: StreamTypes;
+  registration_status: RegistrationStatus;
+  registration_number: number;
+  current_employment_type: CurrentEmploymentTypes;
+  additional_comments?: string;
+
+  // Page 4
+  deploy_anywhere: boolean;
+  deployment_locations: HealthAuthorities[];
+  placement_prefs: PlacementPreferences;
+  is_immunized: boolean;
+  deployment_duration: DeploymentDurations;
+}
+
+// TODO get Stream types
+export enum StreamTypes {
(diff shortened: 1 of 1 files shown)


packages/api/constants/form-structure.ts
```diff
@@ -0,0 +1,85 @@
...
+
+// TODO get values
+export enum MainSpecilaization {
+  SPEC_1 = 'Specialization 1',
+  SPEC_2 = 'Specialization 2',
+  SPEC_3 = 'Specialization 3',
+}
+
+// TODO get values
+export enum SubSpecialization {
```
fwpushan, line 67: Sub-specilaization, could be function/mapped value of MainSpecilaization  `f(mainSpec) -> SunSpec`

packages/api/constants/form-structure.ts
```diff
@@ -0,0 +1,85 @@
+export interface FormData {
+  // Page 1
+  first_name: string;
```
fwpushan, line 3: why snake style is used for property naming?

fwpushan (commented):
Initial Draft looks good. But I've few recommendations 
1. The fields and the expected value range should be clearly mentioned in corresponding ticket. So, will recommend, go with agile approach add what we required for that story.
2.  If there are any cross package usage of these interfaces then declare it in either `common` package or `types` package.

arranfw (commented):
Liking the direction so far, let's chat about this as a team today

hitankar-freshworks:
I think we are going away from the NestJs conventions

hitankar-freshworks:
I find nesting keep things readable. I like the suggest the following structure:
```
FormData {
 personalInformation: {
     firstName: string,
     .
     .
   }
  registrationDetails: {
    streamTypes: string;
   .
  .
  },
  deploymentDetails: {
  deployAnywhere: boolean;
 .
 .
 .
 }
}
```

packages/api/constants/form-structure.ts
hitankar-freshworks: I am curious if it will  work with class-validator if we use generics like `Specialization<Type>`

PR #21, opened 2021-11-29
EHPR-6 | Validation base

arranfw:
This layer will be replaced by some sort of class-validator middle-layer, will work on that once we have some entities made in a common package


packages/web/components/form/validation/primary.tsx
```diff
@@ -0,0 +1,14 @@
+import * as yup from 'yup';
+import { FieldsType } from '.';
```
fwpushan, line 2: Is there any chance of circular module load problem. Because from index.tsx ... primary tsx is imported?

fwpushan (commented):
LGTM, just one comment regarding one clarifications ?

packages/web/components/form/validation/primary.tsx
arranfw: I don't believe so, I've never experience circular dependency issues in Next/CRA

dbayly-freshworks (approved):
This should be ok for now, but If we can use class-validator for validation later, that would be great. Being able to re-use validation between front and back end could be incredibly powerful.

arranfw:
> This should be ok for now, but If we can use class-validator for validation later, that would be great. Being able to re-use validation between front and back end could be incredibly powerful.


Totally agree, will reimplement this section when we get entities created

hitankar-freshworks (approved):

PR #23, opened 2021-11-30
EHPR-47 | web unit tests

arranfw:
- Add unit tests (page + component example) using jest/react-testing-library
- Add web-specific workflow file (should be extended to api/common)
- Move pages/components to `src` folder in web

https://freshworks.atlassian.net/browse/EHPR-47


arranfw:
Before we merge this, we should probably extend the new workflow to the other packages. I think for good efficiency we should split PR checks into web/api/common packages so we don't build/test the whole repo when making a change to a single package

hitankar-freshworks (approved):

vyasworks (approved):
LGTM

PR #24, opened 2021-11-30
Ehpr backend prototype

dbayly-freshworks:
Create scaffold for back end. 
This PR addresses EHPR 59 - Create a form module 
and EHPR 61 - Prototyping DTOs and validation


packages/api/src/form/form.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Injectable } from '@nestjs/common';
+import { Repository } from 'typeorm';
+import { FormEntity } from './entity/form.entity';
+import { FormDTO } from '@ehpr/common/dto/form.dto';
+@Injectable()
+export class FormService {
+  constructor(private readonly formRepository: Repository<FormEntity>) {}
```
hitankar-freshworks, line 7: Dont we need to add `@InjectRepository(FormEntity)`?

packages/api/src/form/form.controller.ts
```diff
@@ -0,0 +1,13 @@
+import { Body, Controller, HttpStatus, Post, Res } from '@nestjs/common';
+import { FormDTO } from '@ehpr/common/dto/form.dto';
+import { FormService } from './form.service';
+
+@Controller('form')
+export class FormController {
+  constructor(private readonly formService: FormService) {}
```
hitankar-freshworks, line 7: missing `@Inject(FormService)`

packages/api/src/form/entity/form.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { FormPayload } from '@ehpr/common/form-payload';
+import { Column, Entity, PrimaryGeneratedColumn } from 'typeorm';
+
+@Entity('form')
+export class FormEntity {
+  @PrimaryGeneratedColumn()
+  id!: number;
+
```
hitankar-freshworks, line 8: Need a reference id column

hitankar-freshworks (approved):
Few comments. Looks good otherwise.

package.json
```diff
@@ -36,5 +36,8 @@
     "tsconfig-paths": "3.10.1",
     "typescript": "4.3.5",
     "wsrun": "5.2.4"
+  },
+  "dependencies": {
+    "class-validator": "^0.13.2"
```
arranfw, line 41: I don't think this belongs in the root package

packages/api/src/form/ro/form.ro.ts
```diff
@@ -0,0 +1,7 @@
+import { FormPayload } from '@ehpr/common/form-payload';
+
+export class FormRO {
+  id!: number;
```
arranfw, line 4: Is this the short, readable reference ID that we'll supply to the user?

packages/common/dto/form.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { IsIn, IsString, ValidateNested } from 'class-validator';
+import { AvailabilityDTO } from './availability-information.dto';
+import { PersonalInformationDTO } from './personal-information.dto';
+import { SkillInformationDTO } from './skill-information.dto';
+
+export class FormDTO {
+  @ValidateNested()
+  payload!: {
```
arranfw, line 8: If possible it'd be nice to sync this naming up with the FE, not sure if we want to add a suffix/prefix but I'm open to changing the FE as long as we can be consistent. Particularly having personal/contact information separate would help

- \<stepName\>Information
- \<stepName\>Page
- etc..

packages/common/dto/placement-preferences.dto.ts
```diff
@@ -0,0 +1,19 @@
+import { IsBoolean, IsOptional } from 'class-validator';
```
arranfw, line 1: Are the validations in these final? If so there are some tickets to update in Jira

packages/common/dto/skill-information.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  streamTypes!: StreamTypes;
+
+  @IsString()
+  @IsIn(Object.values(RegistrationStatus))
+  registrationStatus!: RegistrationStatus;
+
+  @IsNumber()
+  @Max(99999999)
+  @Min(0)
+  registrationNumber!: number;
```
arranfw, line 16: Discussion point: I'd rather represent this as a string if we're not doing any math with it, the min/max is a smell to me I think `IsNumberString` can be used

packages/common/form-payload.ts
```diff
@@ -0,0 +1,88 @@
...
+  personalInformation: {
+    firstName: string;
+    lastName: string;
+    postalCode: string;
+    primaryPhone: string;
+    secondaryPhone: string;
+    email: string;
+  };
+
+  skillInformation: {
```
arranfw, line 11: I would have expected the value of this to be equal to the class `SkillInformationDTO` defined in API, can we get away with reusing it? Is there a reason to duplicate the definition?

arranfw (changes requested):
Let's move the class validator package and then discuss the other comments

packages/common/dto/placement-preferences.dto.ts
dbayly-freshworks: These validations are no final, this is just to have something to work off of.

packages/common/dto/form.dto.ts
hitankar-freshworks: I believe the FE needs to align with the RESTfulness of an API. So, having the terminology contextual is the way6 to go compared to what is being used in the FE.

I don't want us to create a tightly coupled FE + BE just so we can have common entities and classes.

packages/common/dto/skill-information.dto.ts
hitankar-freshworks: Is this the registration number that is autogenerated or something that the user fills in?

dbayly-freshworks: That's totally fine. As stated above, these validators are just a starting point. I figure we'll refine them to be more in line with what makes sense for the front end to send.

package.json
dbayly-freshworks: Aren't we using it for both front and back end validation? Should it be in common instead?

packages/api/src/form/ro/form.ro.ts
dbayly-freshworks: I foresee this as being something we would use in a portal. I don't see it as something the user would interact with.

packages/common/form-payload.ts
dbayly-freshworks: I think it's a question of pattern and use. A DTO in the Nest ecosystem is mean to represent a received object, not the object itself. While we could technically reuse it, it would break nests patterns.

package.json
arranfw: I think each package that needs it should have it installed, packages in root should only be used for monorepo-specific things (lint/prettier/husky) IMO

arranfw: It might only be required in common

packages/common/form-payload.ts
arranfw: Fair enough, I think I'm a bit confused on what exactly the shared parts are meant to be 🤔

arranfw (approved):

PR #25, opened 2021-11-30
EHPR-50 | Add basic docker support for the application

fwpushan:
- [x] Watching of the file in docker development mode
- [ ]  Performance Optimisation  
- [x] Successful local execution
- [x] Single image multiple container 

Ticket: [EHPR-50](https://freshworks.atlassian.net/browse/EHPR-50)

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,8 @@
     "prepare": "husky install",
     "precommit": "run-p 'format:staged'",
     "watch:common": "wsrun --stages --exclude-missing watch",
+    "watch:api": "run-p 'watch:common' 'start:api:local'",
+    "watch:web": "run-p 'watch:common' 'start:web:local'",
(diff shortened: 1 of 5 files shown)


hitankar-freshworks (approved):

package.json
```diff
@@ -14,6 +14,8 @@
     "prepare": "husky install",
     "precommit": "run-p 'format:staged'",
     "watch:common": "wsrun --stages --exclude-missing watch",
+    "watch:api": "run-p 'watch:common' 'start:api:local'",
```
arranfw, line 17: I thought we'd rely on the development tools provided by next/nest rather than using watch, I'm not familiar with a setup like this does it work similarly?

arranfw (approved):

package.json
fwpushan: this is ugly implementation ..... not optimised.

PR #28, opened 2021-12-01
EHPR-57 | Personal info validation

arranfw:
https://freshworks.atlassian.net/browse/EHPR-57

image
image

### DIFF
--- a/packages/common/src/dto/personal-information.dto.ts
+++ b/packages/common/src/dto/personal-information.dto.ts
@@ -1,5 +1,12 @@
 import { IsNotEmpty, IsString, Length, Matches } from 'class-validator';
 
+/**
+ * Canadian postal codes only allow certain sets of characters
+ * This regex helps to validate only those characters are used in the correct position
+ * Also allows a space, dash, or nothing in the middle
+ */
+const postalCodeRegex = /^[ABCEGHJKLMNPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][\s-]?\d[ABCEGHJ-NPRSTV-Z]\d$/i;
+
 export class PersonalInformationDTO {
   @IsString()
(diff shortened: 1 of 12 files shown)


packages/common/src/dto/personal-information.dto.ts
```diff
@@ -1,15 +1,22 @@
...
   @IsString()
-  @Length(1, 255)
+  @Length(1, 255, { message: 'Last Name must be between 1 and 255 characters' })
+  @IsNotEmpty({ message: 'Last Name is Required' })
+  @Matches(/[^ ]+/, { message: 'Last Name is Required' })
   lastName!: string;
 
   @IsString()
-  @Matches(/^[ABCEGHJKLMNPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][\s-]?\d[ABCEGHJ-NPRSTV-Z]\d$/i)
+  @Matches(/^[ABCEGHJKLMNPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][\s-]?\d[ABCEGHJ-NPRSTV-Z]\d$/i, {
```
dbayly-freshworks, line 17: Total nit, but this regex is a bit hard to read. I wonder if it would be worth breaking up/ simplifying. Maybe having a constant for all the valid characters?

dbayly-freshworks (approved):
One nit, otherwise looks good. I'm still blown away by the fact that we're able to use class validator for both front and backend validation

packages/common/src/dto/personal-information.dto.ts
arranfw: Would a link to a source defining it be alright?

dbayly-freshworks: Totally!

hitankar-freshworks (approved):

PR #33, opened 2021-12-02
Ehpr 60 database setup

dbayly-freshworks:
Added database 
Beware, this doesn't build locally right now due to the package shift.


apps/api/src/database/.ormconfig.ts
```diff
@@ -0,0 +1,20 @@
+import * as dotenv from 'dotenv';
+import { ConnectionOptions } from 'typeorm';
+
+dotenv.config();
+
+// Check typeORM documentation for more information.
+export const config: ConnectionOptions = {
```
arranfw, line 7: Is PW the only item we want in env vars? Possibly host/username/db?

apps/api/src/database/database.naming-strategy.ts
```diff
@@ -0,0 +1,47 @@
+import { NamingStrategyInterface, DefaultNamingStrategy } from 'typeorm';
+import { snakeCase } from 'typeorm/util/StringUtils';
+
+export class DatabaseNamingStrategy
```
arranfw, line 4: cool I've never seen this, do the decisions here need to be documented?

apps/api/src/form/form.service.ts
```diff
@@ -14,10 +14,10 @@ export class FormService {
     const newForm = this.formRepository.create(dto);
     return await this.formRepository.save(newForm);
   }
-  async getForms(): Promise<FormEntity[]> {
+  async getForms(): Promise<any> {
```
arranfw, line 17: I'd like to add comments when we add temp code if possible

> This is `any` until ____ happens

docker-compose.yml
```diff
@@ -36,6 +36,8 @@ services:
       - POSTGRES_HOST=db
     depends_on:
       - db
+    networks:
```
arranfw, line 39: Do we need to add a network? Containers should be connected by default I'd be interested to see if it works without a network

apps/api/src/database/.ormconfig.ts
```diff
@@ -0,0 +1,20 @@
...
+  host: 'db',
+  type: 'postgres',
+  port: 5432,
+  username: 'ehpr-admin',
```
hitankar-freshworks, line 11: Might be a good idea to have these values in env file

arranfw:
Is there a migration for that entity missing?

arranfw:
Oops

hitankar-freshworks (approved):

docker-compose.yml
dbayly-freshworks: I added it because I initially had some trouble connecting to the db. I don't think removing this will cause any issues and since I can't get docker to work post folder shift, I might as well remove it. 🤷

dbayly-freshworks:
> Is there a migration for that entity missing?

Right now I've got synchronize on. I'll do migrations once we've established a few more entities and we've deployed something.

apps/api/src/database/database.naming-strategy.ts
dbayly-freshworks: This is part of the FW NestJS boilerplate code. I don't think it's a significant decision/worth documenting.

PR #38, opened 2021-12-02
Ehpr 66 error handling

hitankar-freshworks:
- Error handling and logging with winston
- Interceptors for success and failed responses
- basic swagger setup. payload info is missing but I will add that in a separate PR


apps/api/src/form/form.service.ts
```diff
@@ -14,10 +15,10 @@ export class FormService {
...
     return await this.formRepository.save(newForm);
   }
-  async getForms(): Promise<FormEntity[]> {
+  async getForms(): Promise<any> {
     return await this.formRepository.find();
   }
-  async getFormById(id: number): Promise<FormEntity> {
-    return (await this.formRepository.findByIds([id]))[0];
+  async getFormById(id: number): Promise<any> {
+    // return (await this.formRepository.findByIds([id]))[0];
```
dbayly-freshworks, line 22: It might be worth restoring these if they don't cause any errors.

apps/api/src/form/form.service.ts
```diff
@@ -3,6 +3,7 @@ import { Repository } from 'typeorm';
 import { FormEntity } from './entity/form.entity';
 import { FormDTO } from '@ehpr/common';
 import { InjectRepository } from '@nestjs/typeorm';
+// import { InjectRepository } from '@nestjs/typeorm';
```
dbayly-freshworks, line 6: Commented out export (my bad, I'm guessing this was branched off of one of my WIP branches)

apps/api/src/common/trim.pipe.ts
```diff
@@ -0,0 +1,33 @@
...
+      } else {
+        if (typeof values[key] === 'string') {
+          values[key] = values[key].trim();
+        }
+      }
+    });
+    return values;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dbayly-freshworks, line 23: Called it!  😂 
All good though, having any here makes sense.

apps/api/src/common/documentation.ts
```diff
@@ -0,0 +1,26 @@
+import { INestApplication } from '@nestjs/common';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
+import { FormModule } from 'src/form/form.module';
+
+export const Documentation = (app: INestApplication) => {
+  const options = new DocumentBuilder()
+    .setTitle('eHPR API Documentation')
```
dbayly-freshworks, line 7: I think this is a typo

apps/api/src/common/error-exception.filter.ts
```diff
@@ -0,0 +1,87 @@
...
+   * @param exception: An exception caught by the handler
+   * @returns A failed response object
+   */
+  transformHttpException(exception: Error): FailedResponse {
+    const exceptionMessage: any = exception.message;
+
+    return {
+      errorType:
+        exceptionMessage.error ||
+        (exception as any).response.error ||
```
dbayly-freshworks, line 33: Nit, but is it possible to add a type here?

apps/api/src/common/error-exception.filter.ts
```diff
@@ -0,0 +1,87 @@
...
+      exception instanceof HttpException ? exception.getStatus() : HttpStatus.INTERNAL_SERVER_ERROR;
+    const request = ctx.getRequest();
+    /** Flat error if it was wrapped inside another error */
+    const flattenedException =
+      typeof exception.message === 'object' &&
+      typeof (exception.message as any).message === 'object'
+        ? exception.message
+        : exception;
+
+    const privateKeys: string[] = ['password'];
```
dbayly-freshworks, line 59: Neat! If this list grows, we might want to add it to a constants file.

apps/api/src/form/form.service.ts
```diff
@@ -14,10 +15,10 @@ export class FormService {
     const newForm = this.formRepository.create(dto);
     return await this.formRepository.save(newForm);
   }
-  async getForms(): Promise<FormEntity[]> {
+  async getForms(): Promise<any> {
```
arranfw, line 18: Does this need to be `any`?

apps/api/src/form/form.service.ts
arranfw, line 6: Can we remove this

apps/api/src/common/generic-exception.ts
```diff
@@ -0,0 +1,37 @@
...
+  errorType!: string;
+
+  /** User friendly message */
+  errorMessage!: string;
+
+  /** Default is Internal Server Error */
+  httpStatus?: HttpStatus;
+}
+
+// tslint:disable-next-line: max-classes-per-file
```
dbayly-freshworks, line 16: Nit but we might want to break some of these out into their own files.

apps/api/src/common/documentation.ts
```diff
@@ -0,0 +1,26 @@
+import { INestApplication } from '@nestjs/common';
+import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
+import { FormModule } from 'src/form/form.module';
+
+export const Documentation = (app: INestApplication) => {
```
arranfw, line 5: Why is this in common?

dbayly-freshworks (approved):
Looks ok, but I have a ton of little nits.

arranfw (approved):
Assuming the filters are mostly boilerplate, no major comments 👍

apps/api/src/common/trim.pipe.ts
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Injectable, PipeTransform, ArgumentMetadata } from '@nestjs/common';
+
+@Injectable()
+export class TrimPipe implements PipeTransform {
```
arranfw, line 5: Does this mean FE doesn't need to trim values?

hitankar-freshworks: This is a very nestjs/coupled way of trimming. I don't think it's worth the effort to make it generic

apps/api/src/common/documentation.ts
hitankar-freshworks: Because it is referenced by other nestjs modules. Also, this is common for nestjs API and not for our monorepo 😄

apps/api/src/common/trim.pipe.ts
arranfw: Sorry that was a weird question, does this just trim incoming fields?

apps/api/src/common/documentation.ts
arranfw: Ahh I misread that 👍

hitankar-freshworks:
> Assuming the filters are mostly boilerplate, no major comments 👍

boilerplate now, but was a lot hard work in the past 😄

apps/api/src/common/error-exception.filter.ts
hitankar-freshworks: I really want to get some type for an exception, but it is really hard to be specific as exceptions can be trigger from any part of the stack.

Something we can improve over time?

dbayly-freshworks: Totally! 👍

apps/api/src/common/trim.pipe.ts
hitankar-freshworks: It was a matter of time :-D

apps/api/src/common/trim.pipe.ts
hitankar-freshworks: yeah, its a nice to have. We can remove it if there's no benefit.

PR #42, opened 2021-12-02
EHPR-90 | Radio

arranfw:
https://freshworks.atlassian.net/browse/EHPR-90

- Like a checkbox, but you only get one

image


fwpushan (approved):
LGTM

PR #45, opened 2021-12-06
EHPR-79 | Credential Page UI

arranfw:
https://freshworks.atlassian.net/browse/EHPR-79

Made some changes in `common`, please double check the changes are okay and make sense 👍 

- streamTypes -> stream
- currentEmploymentType -> currentEmployment

I've experienced non-blocking some wierd issues  with React-Select in this implementation, I've spent a bit of time debugging but it feels like weird library stuff, if I had time I would replace it with [downshift](https://github.com/downshift-js/downshift)

image
image

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -30,7 +30,6 @@
     "@testing-library/jest-dom": "5.15.1",
     "@testing-library/react": "12.1.2",
     "@types/react": "17.0.37",
-    "@types/react-select": "5.0.1",
     "autoprefixer": "10.4.0",
     "babel-jest": "27.4.2",
     "eslint": "7.32.0",
--- a/apps/web/src/components/form/MultiSelect.tsx
+++ b/apps/web/src/components/form/MultiSelect.tsx
@@ -50,7 +50,7 @@ export const selectStyleOverride: StylesConfig<OptionType, true> = {
     display: 'flex',
     padding: '1px',
     border: '0',
-    borderBottom: '1px solid #313132',
+    borderBottom: isDisabled ? '2px solid rgb(110, 110, 110)' : '2px solid #313132',
     background: isDisabled ? 'rgb(215, 215, 215)' : '#F5F5F5',
     borderRadius: '0',
   }),
--- a/apps/web/src/components/form/MultiSelect.tsx
+++ b/apps/web/src/components/form/MultiSelect.tsx
@@ -20,14 +20,14 @@ export const MultiSelect: React.FC<MultiSelectProps> = props => {
       label={label}
       component={({ field, form }: FormikFieldProps) => (
         <ReactSelect
+          inputId={name}
+          options={options}
           value={field.value}
           onChange={value => form.setFieldValue(name, value)}
-          inputId={name}
-          isMulti
-          styles={selectStyleOverride}
-          options={options}
-          isDisabled={disabled}
           onBlur={field.onBlur}
+          isDisabled={disabled}
+          styles={selectStyleOverride}
+          isMulti
         />
       )}
     />
@@ -45,13 +45,13 @@ export const selectStyleOverride: StylesConfig<OptionType, true> = {
     transform: 'scale(0.8, 0.85)',
   }),
   input: styles => ({ ...styles }),
-  control: styles => ({
+  control: (styles, { isDisabled }) => ({
     ...styles,
     display: 'flex',
-    padding: '2px',
+    padding: '1px',
     border: '0',
-    borderBottom: '1px solid rgb(49, 49, 50)',
-    background: 'rgb(243, 244, 246)',
+    borderBottom: '1px solid #313132',
+    background: isDisabled ? 'rgb(215, 215, 215)' : '#F5F5F5',
     borderRadius: '0',
   }),
   option: styles => ({ ...styles, padding: '0', paddingLeft: '.5rem' }),
--- a/apps/web/tests/components/submission/components/Credential.test.tsx
+++ b/apps/web/tests/components/submission/components/Credential.test.tsx
@@ -1,0 +1,115 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Credential } from '../../../../src/components/submission/components/Credential';
+import {
+  defaultSpecialtyValue,
+  registrationStatusOptions,
+  healthAuthorityOptions,
+  employmentOptions,
+} from '../../../../src/components/submission/validation/credential';
+import * as formik from 'formik';
+
+const mockSetFieldValue = jest.fn();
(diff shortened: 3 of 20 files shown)


packages/common/src/dto/skill-information.dto.ts
```diff
@@ -1,10 +1,11 @@
-import { CurrentEmploymentTypes, RegistrationStatus, StreamTypes } from '../interfaces';
+import { EmploymentTypes, RegistrationStatus, HealthAuthorities } from '../interfaces';
+import { streamData } from '../data/applicant_stream_data';
 import { IsIn, IsNumber, IsOptional, IsString, Length, Max, Min } from 'class-validator';
 
 export class SkillInformationDTO {
```
arranfw, line 5: I just added the fields I needed here, the decorators aren't final

packages/common/src/data/applicant_stream_data.ts
```diff
@@ -0,0 +1,1057 @@
+export type Stream = Record<
```
fwpushan, line 1: 1. Why file name is in snake case?
2. This file is very big, any thought of refactor this to smaller one?

fwpushan (commented):
LGTM. Check my comments. Apart form that a few clarifications are required 
- I see lot of data source related codes (like Options values) are in validation group (for FrontEnd), any specific reason for that?

arranfw:
> LGTH. Check my comments. Apart form that a few clarifications are required
> 
> * I see lot of data source related codes (like Options values) are in validation group (for FrontEnd), any specific reason for that?

I think I need to rename that folder, it's (almost) become sort of a util/services folder but for functions that are mostly used for the form steps. Validation isn't the right name for it but I like the placement of those files. Do you have a suggestion on a different structure?

packages/common/src/data/applicant_stream_data.ts
arranfw: 1. No good reason 😄 
2. The data can't really be easily split up and I've just added a couple typings for it, can you think of a way to split it up? Maybe back to a JSON file and a separate TS file for types?

fwpushan: Let discuss refactor approach  and feasibility

packages/common/src/dto/skill-information.dto.ts
```diff
@@ -26,11 +27,24 @@ export class SkillInformationDTO {
...
+  healthAuthorities!: HealthAuthorities[];
 
   @IsString()
-  @Length(1, 256)
+  @Length(1, 255)
   @IsOptional()
   additionalComments?: string;
 }
+
+export interface SpecialtyType {
```
arranfw, line 43: Should this be an additional DTO like `PlacementPreferencesDTO`?

apps/web/package.json
```diff
@@ -30,7 +30,6 @@
     "@testing-library/jest-dom": "5.15.1",
     "@testing-library/react": "12.1.2",
     "@types/react": "17.0.37",
-    "@types/react-select": "5.0.1",
```
dbayly-freshworks, line 33: Rip

packages/common/src/dto/skill-information.dto.ts
dbayly-freshworks: I think so. That would match the pattern already established.

apps/web/tests/components/submission/components/Credential.test.tsx
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Credential } from '../../../../src/components/submission/components/Credential';
+import {
+  defaultSpecialtyValue,
+  registrationStatusOptions,
+  healthAuthorityOptions,
+  employmentOptions,
+} from '../../../../src/components/submission/validation/credential';
```
dbayly-freshworks, line 9: Total nit, but we might want to spend a bit of time figuring out pathing for the project.

apps/web/tests/components/submission/components/Credential.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+jest.mock('next/router', () => ({
+  useRouter: () => ({
+    query: {
+      step: '3',
+    },
+  }),
+}));
+
+describe('Credential', () => {
+  it('should render', () => {
```
dbayly-freshworks, line 24: I notice this has a ton of of checks in each it block. Conventional testing good practice suggests having only one test per block to make it easier to track down and fix issues. Whats your take?

apps/web/src/components/form/MultiSelect.tsx
```diff
@@ -45,13 +45,13 @@ export const selectStyleOverride: StylesConfig<OptionType, true> = {
...
+  control: (styles, { isDisabled }) => ({
     ...styles,
     display: 'flex',
-    padding: '2px',
+    padding: '1px',
     border: '0',
-    borderBottom: '1px solid rgb(49, 49, 50)',
-    background: 'rgb(243, 244, 246)',
+    borderBottom: isDisabled ? '2px solid rgb(110, 110, 110)' : '2px solid #313132',
+    background: isDisabled ? 'rgb(215, 215, 215)' : '#F5F5F5',
```
dbayly-freshworks, line 54: Should we add these colours to the theme?

dbayly-freshworks (commented):
I have a ton of small comments here. Overall looks good. Love the testing work that you've done!

apps/web/src/components/form/MultiSelect.tsx
arranfw: Good thinking but we don't really have a global theme object like in MUI, colors are stored in tailwind/it's config. This is a one-off because this library only allows CSSOM for styling afaik. I don't think it's worth the effort of a global theme/color object just for this component

Switching to downshift would allow us to style with tailwind

apps/web/tests/components/submission/components/Credential.test.tsx
arranfw: Jest didn't like the TS absolute pathing, if we are able to fix that configuration it'll look a lot better

apps/web/tests/components/submission/components/Credential.test.tsx
arranfw: I think this comment should be on `should render the expected fields` but yeah I think you're right in general. In this case I am testing for the same thing just for different fields and it felt like it'd be too verbose to write a test for each field. So you're right and the question is "is this shortcut bad enough to fix"

apps/web/package.json
arranfw: Comes by default with v5!

dbayly-freshworks (approved):
All my questions have been addressed. 👍

PR #47, opened 2021-12-06
EHPR-92 | add horizontal padding to mobile form

arranfw:
https://freshworks.atlassian.net/browse/EHPR-92

### Before
image

### After
image

### DIFF
--- a/apps/web/src/components/submission/index.tsx
+++ b/apps/web/src/components/submission/index.tsx
@@ -102,7 +102,7 @@ export const Form: React.FC = () => {
       {({ isSubmitting }) => (
         <FormikForm>
           <div className='flex flex-col items-center'>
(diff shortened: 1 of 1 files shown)


apps/web/src/components/submission/index.tsx
```diff
@@ -78,6 +76,7 @@ export const Form: React.FC = () => {
    * form when the previous step hasn't been completed.
    */
   useEffect(() => {
+    return;
```
dbayly-freshworks, line 79: Wait... why the instant return here?

dbayly-freshworks (approved):

apps/web/src/components/submission/index.tsx
arranfw: bad commit >:( thanks for the catch!

dbayly-freshworks (approved):
Looks good to me 👍

PR #49, opened 2021-12-07
confirmation code + swagger schema

hitankar-freshworks:
Changes include:
- ~~Swagger schema. `nestjs` library for swagger has high dependency on `nestjs` framework. import all the libraries is not worth the simplicity of adding a schema.~~
- Confirmation code. Using a shorter version of UUID with a very low collision rate.
- Added PayloadDTO.
- Added support for entity serialization as a starter.


apps/api/src/form/form.controller.ts
```diff
@@ -12,10 +27,26 @@ export class FormController {
...
+  @ApiBody({
+    schema: SwaggerSchema,
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
   @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
   @HttpCode(HttpStatus.CREATED)
   @Post()
-  async name(@Body() body: FormDTO): Promise<void> {
-    await this.formService.saveForm(body);
+  async name(@Body() body: FormDTO): Promise<FormEntity> {
```
fwpushan, line 37: Do we want to send whole entity as response? I suggest, confirmation id is sufficient

apps/api/src/form/swagger.ts
```diff
@@ -0,0 +1,90 @@
+import {
+  CurrentEmploymentTypes,
+  DeploymentDurations,
+  HealthAuthorities,
+  RegistrationStatus,
+  StreamTypes,
+} from '@ehpr/common';
+import { SchemaObject } from '@nestjs/swagger/dist/interfaces/open-api-spec.interface';
+
+const primaryInformationSchema = {
```
fwpushan, line 10: I saw `ApiProperty` decorator with DTO object to specify schema, to keep single source of truth for schema. Any specific reason of not using that approach ?

fwpushan (commented):
LGTM. Please check my comments.

apps/api/src/form/swagger.ts
arranfw: I agree, I think I'd prefer adding the swagger tags into the common package to reduce duplication

apps/api/src/form/form.controller.ts
arranfw: The entity contained in the RO could theoretically be used to fill info for the "Download a copy" feature (although that data should already be in browser memory at that time)

apps/api/src/form/entity/form.entity.ts
```diff
@@ -1,12 +1,21 @@
...
 @Entity('form')
 export class FormEntity extends BaseEntity {
   @Column('jsonb', { nullable: false })
-  payload!: FormPayload;
+  payload!: PayloadDTO;
+
+  @Column('varchar', { nullable: true })
+  confirmationId!: string;
+
+  @Exclude()
```
arranfw, line 14: Could you explain what the difference is between fields we want to exclude and fields we don't?

apps/api/src/form/id-generator.ts
```diff
@@ -0,0 +1,6 @@
+import { customRandom } from 'nanoid';
+
+export const generateConfirmationId = () =>
+  customRandom('0123456789ABCDEFGHIJKLMNPQRSTUVWXYZ', 13, size => {
```
arranfw, line 4: Possible remove `I` and/or `L` from this list? I ~ 1 ~ L

apps/api/src/form/entity/form.entity.ts
hitankar-freshworks: This is for entity serialization. Only return the properties we want to send back as response.

apps/api/src/form/form.controller.ts
hitankar-freshworks: I actually spent a good deal of time using nestjs decorators for Swagger. However, it's heavily coupled with the framework to the point I had a lot of dependencies. Also, putting them in devDependencies did not work at all even if I had all the necessary dependencies.
Also, a common package should only include functionality used by both app and API. I would argue against putting swagger schema since it is not being used in the FE.

apps/api/src/form/swagger.ts
hitankar-freshworks: I actually spent a good deal of time using nestjs decorators for Swagger. However, it's heavily coupled with the framework to the point I had a lot of dependencies. Also, putting them in devDependencies did not work at all even if I had all the necessary dependencies.
Also, a common package should only include functionality used by both app and API. I would argue against putting swagger schema since it is not being used in the FE.

arranfw: If it'll be too heavyweight to include Swagger in common, let's leave it here

Can we make properties extend types for each form step to make sure they don't go out of sync?

apps/api/src/form/id-generator.ts
hitankar-freshworks: These are all caps. L shouldn't be an issue. One risk though is reducing the options will increase the chance of collision.

apps/api/src/form/form.controller.ts
hitankar-freshworks: Check line 33. We can use a `ClassSerializer` to limit what can be returned. This can be perfected later.

apps/api/src/form/swagger.ts
hitankar-freshworks: @arranfw not sure I fully understand.

arranfw: Was thinking of adding types to the properties field so changes in the DTO would throw TS errors here if it isn't updated

arranfw (approved):

apps/api/src/form/swagger.ts
hitankar-freshworks: Lets revisit this later.

arranfw (approved):

packages/common/src/dto/payload.dto.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  PersonalInformationDTO,
+  ContactInformationDTO,
+  SkillInformationDTO,
+  AvailabilityDTO,
+} from '.';
+
+export class PayloadDTO {
```
dbayly-freshworks, line 8: In my nested validator PR I've created a very similar class. I will combine the two after this merges.

dbayly-freshworks (approved):
Looks great to me.

packages/common/src/dto/payload.dto.ts
hitankar-freshworks: sorry, I knew this will conflict with yours but the submission was not working without this class

PR #50, opened 2021-12-07
Added custom decorator to fix validate nested issue

dbayly-freshworks:
Added a custom validation decorator to validate nested objects. 

It instantiates a new class of the `FormClass`, witch has all the validation methods in it. It then passes that object 
to a validator.validate function. 

WIP in that I haven't figured out how to get it to return the errors object. 
Please let me know what you think.


apps/api/src/form/entity/form.entity.ts
```diff
@@ -7,6 +7,6 @@ export class FormEntity extends BaseEntity {
   @Column('jsonb', { nullable: false })
   payload!: FormPayload;
 
-  @Column('varchar', { nullable: false })
+  @Column('text', { nullable: false })
```
hitankar-freshworks, line 10: I think we must use varchar over text since the content size can be variable. the `text` type takes up fixed space and that is not necessary.

packages/common/src/dto/availability-information.dto.ts
```diff
@@ -1,13 +1,20 @@
-import { IsBoolean, IsIn, IsNumber, Length, ValidateNested } from 'class-validator';
+import { IsArray, IsBoolean, IsIn, IsNumber, ValidateNested } from 'class-validator';
 
 import { DeploymentDurations, HealthAuthorities } from '../interfaces';
 import { PlacementPreferencesDTO } from '.';
 
 export class AvailabilityDTO {
+  constructor(base: AvailabilityDTO) {
```
hitankar-freshworks, line 7: we can use 'plainToClass' method from 'class-transformer' for this for assignment and skip having a constructor.  Use `Object.assign`

packages/common/src/dto/contact-information.dto.ts
```diff
@@ -4,6 +4,13 @@ const TEN_DIGIT_PHONE_REGEX = /^\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]
 const OPTIONAL_TEN_DIGIT_PHONE_REGEX = /^\(?([0-9]{3})\)?[-.●]?([0-9]{3})[-.●]?([0-9]{4})$/;
 
 export class ContactInformationDTO {
+  constructor(base: ContactInformationDTO) {
```
hitankar-freshworks, line 7: please check my previous comment

hitankar-freshworks (approved):
LGTM. a couple of suggestions.

packages/common/src/dto/custom-decorator.ts
```diff
@@ -0,0 +1,20 @@
+import { registerDecorator, ValidationArguments, Validator } from 'class-validator';
+import { FormClass } from './form-class.dto';
+
+const validator = new Validator();
+export function ValidateNestedObject() {
```
arranfw, line 5: is `custom-decorator` the right name for this file? It's written specifically for the Form class so something like `FormValidator` might be more appropritate

packages/common/src/dto/form-class.dto.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  PersonalInformationDTO,
+  ContactInformationDTO,
+  SkillInformationDTO,
+  AvailabilityDTO,
+} from '.';
+import { Type } from 'class-transformer';
+import { IsNotEmpty, ValidateNested } from 'class-validator';
+
+export class FormClass {
```
arranfw, line 10: FormClass doesn't seem right as a name, FormPayload?

packages/common/src/dto/placement-preferences.dto.ts
```diff
@@ -1,6 +1,13 @@
 import { IsBoolean, IsOptional } from 'class-validator';
 
 export class PlacementPreferencesDTO {
```
arranfw, line 3: Should sub-objects be DTOs as well? I like it, we might need to update the specialty objects in 3rd step to be more like this

packages/common/src/dto/form-class.dto.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  PersonalInformationDTO,
+  ContactInformationDTO,
+  SkillInformationDTO,
+  AvailabilityDTO,
+} from '.';
+import { Type } from 'class-transformer';
+import { IsNotEmpty, ValidateNested } from 'class-validator';
+
+export class FormPayload {
```
arranfw, line 10: Looks like the file is still named `form-class`

arranfw (approved):

PR #51, opened 2021-12-07
EHPR-80 | fix form shape

arranfw:
Requirement for EHPR-80 as well as general form submission

- Pages were previously storing flat values instead of being nested like the DTOs
-  `handleValidate` transforms errors into a nested shape to be compatible with nested 
- Use FormikError for Error component to make CheckBoxArray/Radio easier to work with (TIL)


arranfw:
- Updated Error tests to account for Formik tests
- Remove error from CheckboxArray for now, it'll be readded in the 3rd step PR to keep this one a bit smaller

hitankar-freshworks (approved):

PR #52, opened 2021-12-07
EHPR-94 | Setup remote db

fwpushan:
-----------

- [x]  Setup remote VPC + Security groups
- [x]  Setup rds cluster
- [x]  Inject env details for rds cluster to api
- [x]  Add jest capability to common
- [x]  Test infra code
Ticket: [EHPR-94](https://freshworks.atlassian.net/browse/EHPR-94)

### DIFF
--- a/packages/common/src/dto/skill-information.dto.spec.ts
+++ b/packages/common/src/dto/skill-information.dto.spec.ts
@@ -1,0 +1,3 @@
+describe('Test Skill-Information DTO', () => {
+  it('test should pass', () => {});
+});
--- a/terraform/db.tf
+++ b/terraform/db.tf
@@ -16,7 +16,7 @@ resource "aws_rds_cluster_instance" "pgsql" {
 resource "aws_rds_cluster" "pgsql" {
   cluster_identifier  = local.db_name
   engine              = "aurora-postgresql"
-  availability_zones  = var.azs
+  availability_zones  = ["ca-central-1a"]
   database_name       = replace(var.project_code, "-", "_")
   master_username     = var.db_username
   master_password     = data.aws_ssm_parameter.postgres_password.value
--- a/terraform/db.tf
+++ b/terraform/db.tf
@@ -1,0 +1,39 @@
+
+resource "aws_db_subnet_group" "pgsql" {
+  name       = "pgsql"
+  subnet_ids = [aws_subnet.private_subnet.id]
+}
+
+resource "aws_rds_cluster_instance" "pgsql" {
+  count              = var.target_env == "prod" ? 2 : 1
+  identifier         = "${local.db_name}-${count.index}"
+  cluster_identifier = aws_rds_cluster.pgsql.id
+  instance_class     = "db.r5.large"
+  engine             = aws_rds_cluster.pgsql.engine
+  engine_version     = aws_rds_cluster.pgsql.engine_version
+}
+
+resource "aws_rds_cluster" "pgsql" {
+  cluster_identifier  = local.db_name
+  engine              = "aurora-postgresql"
+  availability_zones  = var.azs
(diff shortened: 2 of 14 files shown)


terraform/db.tf
```diff
@@ -0,0 +1,39 @@
...
+resource "aws_db_subnet_group" "pgsql" {
+  name       = "pgsql"
+  subnet_ids = [aws_subnet.private_subnet.id]
+}
+
+resource "aws_rds_cluster_instance" "pgsql" {
+  count              = var.target_env == "prod" ? 2 : 1
+  identifier         = "${local.db_name}-${count.index}"
+  cluster_identifier = aws_rds_cluster.pgsql.id
+  instance_class     = "db.r5.large"
```
dbayly-freshworks, line 11: This seems pretty big. How many records are we expecting?

dbayly-freshworks (approved):
Looks fine to me, one comment.

hitankar-freshworks:
@fwpushan we will need a way to access the db locally. suggest we do what LTC has.

terraform/network.tf
```diff
@@ -0,0 +1,56 @@
+resource "aws_vpc" "vpc" {
```
hitankar-freshworks, line 1: Can we just not just go with a security group and since this is just temporary

terraform/db.tf
fwpushan: This code is copied from ltx and ets ..... Almost or a little bigger than HCAP.

terraform/network.tf
fwpushan: As per @v-y-a-s suggestion. We deployed it our sandboxed env and this will be destroyed.

fwpushan:
> @fwpushan we will need a way to access the db locally. suggest we do what LTC has.

@hitankar-freshworks 
This is the first part of the sub-task and this will unblock other. 
In the next part I will setup bastion host.

terraform/db.tf
vyasworks: Maybe something smaller would do? I highly doubt we need R5 with 16 gb memory. 

Maybe use a general purpose category? https://aws.amazon.com/rds/instance-types/

terraform/network.tf
vyasworks: This might take a bit of time to setup but nevertheless, a good exercise. 

We will be using LZ2's VPC when we get access

arranfw:
> > @fwpushan we will need a way to access the db locally. suggest we do what LTC has.
> 
> @hitankar-freshworks This is the first part of the sub-task and this will unblock other. In the next part I will setup bastion host.

Could you document the tasks in the ticket? https://freshworks.atlassian.net/browse/EHPR-29

packages/common/src/dto/skill-information.dto.spec.ts
```diff
@@ -0,0 +1,3 @@
+describe('Test Skill-Information DTO', () => {
```
vyasworks, line 1: Not sure if this belongs in this PR, but no issues :smile:

terraform/db.tf
```diff
@@ -0,0 +1,39 @@
...
+  engine             = aws_rds_cluster.pgsql.engine
+  engine_version     = aws_rds_cluster.pgsql.engine_version
+}
+
+resource "aws_rds_cluster" "pgsql" {
+  cluster_identifier  = local.db_name
+  engine              = "aurora-postgresql"
```
vyasworks, line 18: I wonder what the difference in postgres and aurora postgres engine would be? 
Docs from aws say just postgres is also an option, tf docs says aurora.. Unsure about this. 

would be good to know the difference. 

https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-cluster.html

PR #53, opened 2021-12-07
Form Submission

arranfw:
This allows us to submit partial form values to the API

**NOTE** You'll either need to run the webapp through docker or add `NEXT_PUBLIC_API_URL=http://localhost:4000/api/v1` to a file called `.env.local` in `apps/web`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,6 +51,7 @@ services:
     ports:
       - '3000:3000'
     environment:
+      - NEXT_PUBLIC_API_URL=http://localhost:4000/api/v1
       - NODE_ENV=development
       - RUNTIME_ENV=local
     depends_on:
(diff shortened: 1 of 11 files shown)


docker-compose.yml
```diff
@@ -51,6 +51,7 @@ services:
     ports:
       - '3000:3000'
     environment:
+      - NEXT_PUBLIC_API_URL=http://localhost:4000/api/v1
```
dbayly-freshworks, line 54: Will this work when deployed? Should we create an environment variable?

arranfw: oh yeah this is going to need to be added to TF, thanks

hitankar-freshworks: This will be dynamic based on what cloudfront sets. I think LTC has a way to set this based on the environment.

hitankar-freshworks (approved):
One change is requested. Dynamically set url in the Makefile and reference that in `docker-compose.yml`.

PR #54, opened 2021-12-08
EHPR-95 | faster local docker dev

arranfw:
https://freshworks.atlassian.net/browse/EHPR-95

Pretty unconventional setup, relies on the host machine to manage node_modules in all packages. This will never be useable for deployment but it works well for local dev, modeled after LTC

The only major deficiency I've encountered is that it takes quite a long time to stop the containers without forcing


apps/api/Dockerfile.dev
```diff
@@ -0,0 +1,7 @@
+FROM node:14.17.1
+
```
hitankar-freshworks, line 2: Filename Dockerfile should be fine.

hitankar-freshworks (approved):

apps/api/Dockerfile.dev
arranfw: I meant for `.dev` to be a marker that this isn't a typical dockerfile and isn't meant to be used for deployments

hitankar-freshworks: I agree as this is is appropriate where we deploy with Docker. In this case, we don't have a prod Dockerfile so, just Sockerfile should be okay

Makefile
```diff
@@ -92,7 +92,8 @@ docker-build:
 
 docker-run:
 	@echo "++\n***** Running docker-compose\n++"
-	@docker-compose up
+	@yarn
+	@docker compose up --build
```
hitankar-freshworks, line 96: While we are at it, I intentionally removed the build param because it is quite slow. Maybe it won't be an issue with your current setup. Ideally, I was thinking we can do another makefile command like `docker-up`  that runs 'docker-build' and `docker-run`.

arranfw: I like that idea!

sunghwan-park-fw (approved):

PR #55, opened 2021-12-08
Fix build

arranfw:
Some types changing in common didn't case a merge conflict but broke the build after PR checks happened 🤔 Curious how we can prevent this in the future

- DeepPartial is temporary while form steps are not fully implemented

### DIFF
--- a/apps/web/src/components/submission/index.tsx
+++ b/apps/web/src/components/submission/index.tsx
@@ -8,6 +8,7 @@ import {
   contactSchema,
   SubmissionType,
   initialSubmissionValues,
+  DeepPartial,
 } from './validation';
 import { Ret } from 'class-validator-formik/dist/convertError';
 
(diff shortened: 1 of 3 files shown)


apps/web/src/components/submission/index.tsx
```diff
@@ -53,8 +54,8 @@ const steps: StepType[] = [
 
 const handleValidate = (
   validator: (data: unknown) => Ret,
-  values?: SubmissionType,
-  key?: keyof SubmissionType,
+  values?: DeepPartial<SubmissionType>,
+  key?: keyof DeepPartial<SubmissionType>,
```
dbayly-freshworks, line 58: What is the difference between a Partial and a DeepPartial?

dbayly-freshworks (approved):
Looks fine to me!

apps/web/src/components/submission/index.tsx
arranfw: Almost the same thing as ValidateNested

`Partial<SubmissionType>` doesn't like when `personalInformation.firstName` can be undefined

PR #58, opened 2021-12-08
EHPR-95 | Zero install Docker

fwpushan:
- [x] Fix web build issue
- [x] Standalone container to watch common changes 
- [x] Attach common changes with api and web container.
- [x] Create uniform cmd interface for all containers. 
- [x] Fix some file copy issue related to aws api build zip

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,10 @@ FROM node:16-alpine
 
 # RUN yarn set version berry
 # Copying repo resources
-COPY ./package.json ./
-COPY ./tsconfig.json ./
 COPY ./packages ./packages
-COPY ./nodemon.json ./
-COPY ./.eslintrc.js ./
 COPY ./apps ./apps
-RUN yarn install
-
-# Build whole all packages
-RUN yarn build
+COPY ./package.json ./package.json
+COPY ./tsconfig.json ./tsconfig.json
 
 EXPOSE 3000
 EXPOSE 4000
--- a/apps/web/.babelrc
+++ b/apps/web/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["next/babel"]
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,10 +23,9 @@ services:
       dockerfile: Dockerfile
     command: 'yarn watch:api'
     volumes:
+      - ./node_modules:/node_modules
       - ./apps/api:/apps/api
-      - '/apps/api/node_modules'
-      - ./packages/common:/packages/common
-      - '/packages/common/node_modules'
+      - ./packages:/packages
     ports:
       - '4000:4000'
     env_file:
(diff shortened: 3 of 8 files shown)


docker-compose.yml
```diff
@@ -42,12 +41,11 @@ services:
     build:
       context: ./
       dockerfile: Dockerfile
-    command: 'yarn watch:web'
+    command: 'yarn workspace @ehpr/web start'
```
arranfw, line 44: I'm not a huge fan of the inconsistency in commands here

arranfw (commented):
It would be nice to have things a bit more organized, I liked the idea of having a container running yarn watch for the common package, it doesn't totally make sense why api is watching common and web isn't

Dockerfile
```diff
@@ -2,16 +2,10 @@ FROM node:16-alpine
 
 # RUN yarn set version berry
 # Copying repo resources
-COPY ./package.json ./
-COPY ./tsconfig.json ./
 COPY ./packages ./packages
-COPY ./nodemon.json ./
-COPY ./.eslintrc.js ./
 COPY ./apps ./apps
```
sunghwan-park-fw, line 6: I am not sure if this copy command is consistent with the volume mappings in docker-compose.yml.

docker-compose.yml
fwpushan: FIXED

fwpushan: Fixed

fwpushan:
> It would be nice to have things a bit more organized, I liked the idea of having a container running yarn watch for the common package, it doesn't totally make sense why api is watching common and web isn't

@arranfw Implemented a standalone watch container for common and attach them with api and web. Everything is running very smooth in my local host.

Dockerfile
fwpushan: COPY cmd copy asset to image (static binding) build phase.
VOLUME mounting is attaching local file system with running container (Dynamic).

dbayly-freshworks (approved):
I just pulled this down locally and it works great. Good job!

Dockerfile
```diff
@@ -2,16 +2,10 @@ FROM node:16-alpine
 
 # RUN yarn set version berry
 # Copying repo resources
-COPY ./package.json ./
-COPY ./tsconfig.json ./
 COPY ./packages ./packages
```
hitankar-freshworks, line 5: Do we need to even copy packages and apps if we are mounting them?

packages/common/src/hello-world.ts
```diff
@@ -1 +1 @@
-export const helloWorld = () => 'Hello World EHPR!!! :)';
+export const helloWorld = () => 'Hello World EHPR!!!';
```
hitankar-freshworks, line 1: we can remove this file.

Dockerfile
arranfw: I think we shouldn't need to copy if we mount directory as a volume

hitankar-freshworks:
LGTM, one comment

hitankar-freshworks (approved):

apps/web/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["next/babel"]
```
arranfw, line 2: Why did this need to be added?

sunghwan-park-fw: maybe https://nextjs.org/docs/messages/failed-loading-swc

fwpushan: web building is failing in docker env.

sunghwan-park-fw: I guess it is because SWC is a binary module specific to the platform and we don't install node modules in the container.

PR #59, opened 2021-12-09
[EHPR-9] CSV Export

vyasworks:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
fwpushan, line 2: What is the significant of this setting?

fwpushan (approved):
LGTM

apps/api/src/form/form.service.ts
```diff
@@ -2,7 +2,8 @@ import { Injectable } from '@nestjs/common';
 import { Repository } from 'typeorm';
 import { FormEntity } from './entity/form.entity';
```
hitankar-freshworks, line 3: I wish we have ordering of imports with third party packages first then our codebase.

apps/api/src/form/form.service.ts
```diff
@@ -26,4 +27,45 @@ export class FormService {
...
   }
+
+  async getAllForms(): Promise<FormEntity[]> {
+    return await this.formRepository.find({});
+  }
+
+  async flattenAndTransformFormData(forms: FormEntity[]): Promise<FormExportColumns[]> {
+    return forms.map(form => {
+      return {
+        deployAnywhere: booleanToYesNo(form.payload.availabilityInformation.deployAnywhere),
```
hitankar-freshworks, line 38: it might look better to destructure `form` to get `payload`.

apps/api/src/helpers/csv.ts
```diff
@@ -0,0 +1,22 @@
+import { stringify } from 'csv-stringify';
```
hitankar-freshworks, line 1: Nest is modular, so it might be a good idea to put it under `src/common`. Also, `helpers` could be singular

packages/common/src/interfaces/formExport.ts
```diff
@@ -0,0 +1,66 @@
+// Order determines the CSV column order
```
hitankar-freshworks, line 1: This file can be in the api only. FE doesn't need it.

packages/common/src/interfaces/index.ts
```diff
@@ -1 +1,2 @@
 export * from './form-payload';
+export * from './formExport';
```
hitankar-freshworks, line 2:
```suggestion
export * from './form-export';
```
Keep naming consistent

terraform/api.tf
```diff
@@ -16,12 +16,13 @@ resource "aws_lambda_function" "api" {
...
     variables = {
-      NODE_ENV           = var.target_env
-      AWS_S3_REGION      = var.region
-      RUNTIME_ENV        = "hosted"
-      POSTGRES_HOST      = aws_rds_cluster.pgsql.endpoint
-      POSTGRES_DATABASE  = aws_rds_cluster.pgsql.database_name
-      POSTGRES_PASSWORD  = data.aws_ssm_parameter.postgres_password.value
+      NODE_ENV          = var.target_env
+      AWS_S3_REGION     = var.region
+      RUNTIME_ENV       = "hosted"
```
hitankar-freshworks, line 21: what is the benefit of setting `hosted`?

hitankar-freshworks (commented):
Functionality is good. Few inconsistencies in naming and placing of new files

terraform/api.tf
vyasworks: Not sure about that, I have added only the `export` secret.

packages/common/src/interfaces/formExport.ts
vyasworks: moved

.vscode/settings.json
vyasworks: so that VS code uses the Typescript version in the root `package.json` and not any other version installed in your system

apps/api/src/form/form.service.ts
vyasworks: I think this might work, https://github.com/import-js/eslint-plugin-import

or https://github.com/trivago/prettier-plugin-sort-imports

but let's get it done separately.

apps/api/src/form/form.service.ts
vyasworks: done

terraform/api.tf
vyasworks: the diff is because of the reformat

hitankar-freshworks (approved):
LGTM, 1 small change request regarding swagger.

apps/api/src/common/helper/csv/casting.ts
```diff
@@ -0,0 +1,3 @@
+export const booleanToYesNo = (bool: boolean) => {
```
arranfw, line 1: Could these kinds of generic helpers go in `packages/common`?

apps/api/src/form/form.controller.ts
```diff
@@ -43,4 +50,39 @@ export class FormController {
...
     return await this.formService.getFormById(id);
   }
+
+  @ApiOperation({
+    summary: 'Export Data',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
+  @HttpCode(HttpStatus.CREATED)
+  @Get('/export/:passCode')
```
arranfw, line 60: I think passcode should be in a header rather than a query param, no?

apps/api/src/form/form.controller.ts
```diff
@@ -43,4 +50,39 @@ export class FormController {
...
+  async exportAll(@Param('passCode') passCode: string, @Res() res: Response) {
```
arranfw, line 61: I think this is more logic than I'd prefer in a controller, can this be made into a service file?

apps/api/src/form/form.controller.ts
```diff
@@ -43,4 +50,39 @@ export class FormController {
...
+
+    // TODO: Based on auth roles in the future.
+    if (passCode === process.env.EXPORT_SECRET) {
+      res.set({
+        'Content-Type': 'text/csv',
+        'Content-Disposition': 'attachment;filename=ehrp.csv',
+      });
+
+      // NOTE: Could be paginated
+      // Current Metric ~ 150K records ~ 3 sec ~ 38 MB
```
arranfw, line 72: Napkin math means we can support 23k records before hitting the 6mb lambda payload limit? Should be fine for now

apps/api/src/form/form.controller.ts
```diff
@@ -43,4 +50,39 @@ export class FormController {
...
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
+  @HttpCode(HttpStatus.CREATED)
+  @Get('/export/:passCode')
+  async exportAll(@Param('passCode') passCode: string, @Res() res: Response) {
+    // TODO: Add logging
+
+    // TODO: Based on auth roles in the future.
+    if (passCode === process.env.EXPORT_SECRET) {
```
arranfw, line 65: We can save some nesting by throwing early rather than surrounding the whole block in an if statement 

```
    if (passCode !== process.env.EXPORT_SECRET) {
        throw new....
    }
```

arranfw (changes requested):

apps/api/src/form/form.controller.ts
vyasworks: No. the end users cant make a API call by setting headers. We also cannot share a url if we do this. That is the reason.

apps/api/src/form/form.controller.ts
vyasworks: That's the payload limit - NOT the download to client limit. We have no payload in this case.

apps/api/src/form/form.controller.ts
vyasworks: I don't think so. I initially had a csv generator module, and it's very little functionality to have it's own service. 

Also we need to do the filtering and sorting here anyways.

apps/api/src/form/form.controller.ts
```diff
@@ -43,4 +50,39 @@ export class FormController {
   async getFormById(@Param('id') id: string): Promise<FormEntity> {
     return await this.formService.getFormById(id);
   }
+
+  @ApiOperation({
+    summary: 'Export Data',
+  })
+  @UseInterceptors(ClassSerializerInterceptor)
+  @ApiResponse({ status: HttpStatus.CREATED, type: EmptyResponse })
```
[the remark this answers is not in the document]
vyasworks: I was thinking of the same, We dont have a response type.. it is a stream file so I made it empty reponse

apps/api/src/form/form.controller.ts
arranfw: Interesting, I thought we'd faced issues with that in the past, I could be wrong

vyasworks: image

apps/api/src/form/form.controller.ts
arranfw: Who are the end users of this endpoint? I assumed it was just us

I can't logically think of a reason I prefer a header (other than tracking in a browser's history), if you have a preference for a query param then that's alright

vyasworks: Assumption - 

MVP: 
we share the URL which has the passcode to MoH user. 
MoH User downloads a copy whenever they want 

> This way, there needs to be no back and forth and they can generate as many as often as required 

Later: 

Have a login + UI + etc.

apps/api/src/form/form.controller.ts
arranfw: Maybe I'm confused, why isn't it a payload?

vyasworks: AH! that's bad. Okay, I think still it's fine, until we convert to storing this into S3. 

Napkin math works for now !

vyasworks: I meant request has no payload.

arranfw: Yep sounds good, should be a while before we hit 6mb. I'll add a ticket to the backlog

PR #62, opened 2021-12-09
EHPR-71 Email module

hitankar-freshworks:
Draft PR to get early feedback

- [x] email module.
- [x] way to add different emails with their templates in the future.
- [x] 2 ways to send email CHES or Mailtrap.
- [x] endpoint to test if email service is working. we are protect this or remove it in the near future .
- [x] templates are currently broken. WIP.


arranfw:
I'm confused about the mailtrap/nodemailer implementation, I thought we were pretty locked into using CHES why are we supporting two methods?

fwpushan:
@hitankar-freshworks @arranfw @dbayly-freshworks 
I don't see any issue of using Mailtrap but my biggest concerns are 
- I always advocate same tech stack and identical environment for all three deployment env (dev/test/prod). So, using different approach and service may create quality and performance issue.
- Security concern, in our STRA, we have to explicitly mention Mailtrap. That, can create some confusion in future.
- Testing, consider functionality is working in dev/test env but not in prod. So, this approach may create confusion and difficulties for testing.
- Effort duplication, if we are using two services for same job, then we have to maintain two code base with conditional branching.
N.B. May be I misunderstood, but my idea was after yesterdays discussion that we are considering Mailtrap as alternative of CHESS as better service.

hitankar-freshworks:
@arranfw @fwpushan removed nodemailer and Mailtrap stuff. Service is fully functional

fwpushan (approved):
LGTM 👍 . Please check why deployment is failing, possibly no ssm param found.

terraform/api.tf
```diff
@@ -16,14 +16,19 @@ resource "aws_lambda_function" "api" {
...
     variables = {
-      NODE_ENV          = var.target_env
-      AWS_S3_REGION     = var.region
-      RUNTIME_ENV       = "hosted"
-      POSTGRES_HOST     = aws_rds_cluster.pgsql.endpoint
-      POSTGRES_DATABASE = aws_rds_cluster.pgsql.database_name
-      POSTGRES_PASSWORD = data.aws_ssm_parameter.postgres_password.value
-      EXPORT_SECRET     = data.aws_ssm_parameter.export_key.value
-      POSTGRES_USERNAME = var.db_username
+      NODE_ENV           = var.target_env
```
dbayly-freshworks, line 19: Nit, but if we want to trick which environment the application is running in, we should use a different variable. NODE_ENV works is unintuitive ways.

dbayly-freshworks (approved):
Looks fine to me! One nit

Makefile
```diff
@@ -11,6 +11,8 @@ export PROJECT := $(or $(PROJECT),ehpr)
 export NODE_ENV ?= development
 export ENV_NAME ?= dev
 export POSTGRES_USERNAME = freshworks
+export CHES_CLIENT_ID ?= EHPR_SERVICE_CLIENT
+export MAIL_FROM ?= noreply@gov.bc.ca
```
arranfw, line 15: `EHPRDoNotReply@gov.bc.ca` is the `from` documented in the EHPR text document, should we use that here instead?

apps/api/src/form/entity/form.entity.ts
```diff
@@ -5,15 +5,16 @@ import { FormPayloadDTO } from '@ehpr/common';
...
   payload!: FormPayloadDTO;
 
   @Column('varchar', { nullable: true })
   confirmationId!: string;
 
   @Exclude()
-  @Column('timestamp', { nullable: true })
-  notifiedAt?: Date;
+  @Column('varchar', { nullable: true })
+  chesId?: string;
```
arranfw, line 17: Determined we don't need a `notifiedAt` column? I don't think it was a requirement right>

terraform/api.tf
hitankar-freshworks: Not sure I fully understand.

apps/api/src/form/form.service.ts
```diff
@@ -1,25 +1,55 @@
...
 
-    return await this.formRepository.save(newForm);
+    const savedForm = await this.formRepository.save(newForm);
+
+    const notifiedForm = await this.sendMail(savedForm);
+
+    return notifiedForm;
+  }
+
+  private async sendMail(form: FormEntity) {
```
arranfw, line 34: `sendMail` seems too generic and sounds like it should be in the mail module, does this function need to be moved or should it be `sendFormConfirmation[Email]`?

apps/api/src/form/entity/form.entity.ts
arranfw: would `chesTransactionId` be better?

apps/api/src/mail/mail.module.ts
```diff
@@ -0,0 +1,11 @@
+import { Module } from '@nestjs/common';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { FormEntity } from 'src/form/entity/form.entity';
+import { MailService } from './mail.service';
+
+@Module({
+  imports: [TypeOrmModule.forFeature([FormEntity])],
```
arranfw, line 7: Nest noob here, does this mean MailModule uses Typeorm? I don't see it used in the service file

hitankar-freshworks:
sorry @arranfw I had two approvals. already merged my PR. I will take notes of your comments and address them in my next PR.

apps/api/src/mail/mail.module.ts
hitankar-freshworks: Good catch. We are not using the entity in the mail. I'll remove it.

apps/api/src/form/form.service.ts
hitankar-freshworks: Good idea

apps/api/src/form/entity/form.entity.ts
hitankar-freshworks: I figured saving timestamp is pointless as both actions happen at the same time. And yes `chesTransactionId` would be better

Makefile
hitankar-freshworks: I think we probably should

PR #66, opened 2021-12-10
Add terraform PR check

arranfw:
Changes to TF directory run all deployment steps except for actually deploying


dbayly-freshworks (approved):
Looks fine, the only potential issue I see is that we may not be able to get static AWS keys for LZ2. Thats a future problem though.

arranfw:
> Looks fine, the only potential issue I see is that we may not be able to get static AWS keys for LZ2. Thats a future problem though.

Interesting, @fwpushan are we aware of that risk?

PR #67, opened 2021-12-10
EHPR-93 fix optional field validation error

sunghwan-park-fw:
Refer to [EHPR-93](https://freshworks.atlassian.net/browse/EHPR-93)


arranfw (approved):
Both regexes were the same 🤔 oops!

So we're replacing the nullish check in `IsOptional` with a falsey check in `ValidateIf` seems like it should work

arranfw:
Let's link up to set up prettier on your machine to fix formatting errors as you develop

dbayly-freshworks (approved):
Looks great! 👍

PR #68, opened 2021-12-10
EHPR-25 | Create single config for nest app of both lambda and other env

fwpushan:
- [x] Create single nest + express app config
- [x] Use same app in aws lambda server-less handler 
- [x] Tested in local
- [ ] Tested in aws

### DIFF
--- a/apps/api/src/lambda.ts
+++ b/apps/api/src/lambda.ts
@@ -1,6 +1,4 @@
-import { NestFactory } from '@nestjs/core';
 import serverlessExpress from '@vendia/serverless-express';
-import express from 'express';
 import {
   APIGatewayProxyEvent,
   APIGatewayProxyResult,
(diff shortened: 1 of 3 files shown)


arranfw (approved):

apps/api/src/lambda.ts
```diff
@@ -1,32 +1,21 @@
...
-    const nestApp = await NestFactory.create(AppModule, new ExpressAdapter(expressApp));
-
-    nestApp.setGlobalPrefix(API_PREFIX);
-
+    const { app: nestApp, expressApp } = await createNestApp();
     await nestApp.init();
-
     cachedServer = serverlessExpress({ app: expressApp });
   }
-
```
hitankar-freshworks, line 29: @fwpushan have you tested if this is working locally?

hitankar-freshworks (approved):
Left one comment.

apps/api/src/lambda.ts
fwpushan: regular app is working locally.

PR #69, opened 2021-12-10
EHPR-80 | Third step validation

arranfw:
- `SkillInformation` validation now deeply validates stream/spec/subspecs using the data file. 
- Renamed subSpecialty -> subspecialty everwhere, this was causing confusion
- added `employmentCircumstance` question as per design update
- Converted applicant stream data to normalized shape and added some ORM-like methods to retrieve it's data
- FE methods for populating streams updated and simplified


fwpushan (approved):
LGTM 👍

hitankar-freshworks:
I maybe be overly cautious here but FE and BE are sharing validation. Although I am not saying this is not happening, any structural changes that improve the FE validation needs checking to make sure BE validation is updated. 

The payload sent to the API endpoint can be used and altered even though the FE validates. Any lapse in revalidation at the API level can be risky.

fwpushan (approved):
LGTM

packages/common/src/dto/is-array-of-specialties.decorator.ts
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const subspecialtyNotListed = (subspecialty: SubspecialtyDTO, formSpecialties: Subspecialty[]) => {
+  if (!formSpecialties.find(formSpecialty => formSpecialty.id === subspecialty.id)) {
+    return true;
+  }
+};
+
+@ValidatorConstraint({ name: 'specialties', async: false })
+export class IsArrayOfSpecialties implements ValidatorConstraintInterface {
```
arranfw, line 43: I'm not sure if there is a way to simplify this at all, this is where the majority of the debt from this ticket lays

PR #70, opened 2021-12-11
EHPR-105 scroll to the top of the form

sunghwan-park-fw:
- update header text
- scroll to the top of the form at step transitions
- do not scroll at initial loading
- rollback logo to img element
[exportPathMap](https://nextjs.org/docs/api-reference/next.config.js/exportPathMap) should be defined in `next.config.js` in order to benefit from `image optimization`. But it can't be available because of dynamic paths.

### DIFF
--- a/apps/web/src/pages/submission/[step].tsx
+++ b/apps/web/src/pages/submission/[step].tsx
@@ -53,7 +53,7 @@ const Submission = () => {
           please have your registration number available.
         </p>
         <p className='mb-4'>
-          <b>If you encounter problems completing/submitting this form, please email </b>
+          <b>If you encounter problems completing this form, please email </b>
           <ExternalLink href='mailto:EHPRQuestions@gov.bc.ca'>EHPRQuestions@gov.bc.ca</ExternalLink>
           .
         </p>
--- a/apps/web/src/pages/submission/[step].tsx
+++ b/apps/web/src/pages/submission/[step].tsx
@@ -1,6 +1,7 @@
 import { useRouter } from 'next/router';
 
 import { ExternalLink, Form, Stepper } from '@components';
+import { useEffect, useRef } from 'react';
 
 const FORM_STEPS = ['Primary', 'Contact', 'Credential', 'Preferences', 'Review and Submit'];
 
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/submission/[step].tsx
```diff
@@ -9,34 +10,70 @@ const Submission = () => {
 
   const step = Number(router.query.step);
 
+  // do not scroll to the top of the form on initial load
+  const formRef = useRef<HTMLDivElement>(null);
+  const prevStepRef = useRef<number>(1);
+  useEffect(() => {
```
Comment (arranfw) [inline apps/web/src/pages/submission/[step].tsx:16]: Would be awesome to have this extracted in to a hook that takes a dependency array and returns a reference to a ref, both for reusability and to reduce component logic

apps/web/src/pages/submission/[step].tsx
```diff
@@ -9,34 +10,70 @@ const Submission = () => {
...
+    if (!isNaN(step) && prevStepRef.current !== step && formRef.current) {
+      window.scrollTo({ left: 0, top: formRef.current.offsetTop, behavior: 'smooth' });
```
Comment (arranfw) [inline apps/web/src/pages/submission/[step].tsx:18]: I'm 50/50 on the use of `behavior: 'smooth'`, it works well in some cases but in others it results in an instant scroll to the top of the page and then a smooth scroll downward toward the form which I don't think is the best behaviour

arranfw (commented):
Great solution, works very well, a couple comments but otherwise this should work great

apps/web/src/pages/submission/[step].tsx
```diff
@@ -9,34 +10,70 @@ const Submission = () => {
...
         </p>
-        <p>
-          If you encounter problems completing/submitting this form, please email{' '}
+        <p className='mb-4'>
+          The form takes about 10 minutes to complete - if you are in a career stream with oversight
+          from a regulatory or credentialing body (e.g., BC College of Nursing Professionals),
+          please have your registration number available.
+        </p>
+        <p className='mb-4'>
+          <b>If you encounter problems completing/submitting this form, please email </b>
```
Comment (arranfw) [inline apps/web/src/pages/submission/[step].tsx:56]: > If you encounter problems completing this form, please email EHPRQuestions@gov.bc.ca.

The text from the doc only says `completing` rather than `completing/submitting`, do we have different sources for copy?

Comment (sunghwan-park-fw) [reply apps/web/src/pages/submission/[step].tsx:56]: My mistake.

hitankar-freshworks (approved):

arranfw (approved):

PR #72, opened 2021-12-14
EHPR-108 remove BE validation

hitankar-freshworks:
- BE validation removed temporarily
- Changes needed to run e2e locally
- Changes to tackle @arranfw 's comment from my last PR


hitankar-freshworks:
> I would prefer the validation removal PR to not have additional things added to it, so that we can revert it later

Makes sense. Let me do a separate PR for this

PR #74, opened 2021-12-15
EHPR-108 | Fix DB connection issue

fwpushan:
- [x] Uncouple nest application config code from bootstrap code


arranfw (approved):
Interesting, just changing the file fixes the issue? Could you explain what the problem was?

hitankar-freshworks:
I'd be surprised if changing the file would work. I was thinking in the line of our TF config. Maybe VPC?

arranfw (changes requested):
If the file main->config file change isn't needed lets revert it

sunghwan-park-fw (approved):

apps/api/src/.ormconfig.ts
```diff
@@ -4,7 +4,6 @@ import { DatabaseNamingStrategy } from './database/database.naming-strategy';
 dotenv.config();
 // Check typeORM documentation for more information.
 export const config: ConnectionOptions = {
-  name: 'default',
```
hitankar-freshworks, line 7: I don't think this is the issue. perhaps we can set `connectTimeoutMS` to `10000`

apps/api/src/.ormconfig.ts
```diff
@@ -4,7 +4,6 @@ import { DatabaseNamingStrategy } from './database/database.naming-strategy';
 dotenv.config();
 // Check typeORM documentation for more information.
 export const config: ConnectionOptions = {
```
hitankar-freshworks, line 6:
```suggestion
export const config: PostgresConnectionOptions = {
```

apps/api/src/main.ts
```diff
@@ -1,92 +1,4 @@
...
-
-  return {
-    app,
-    expressApp,
-  };
-}
+import { createNestApp } from './app.config';
 
 async function bootstrap() {
   const { app, ...other } = await createNestApp();
```
hitankar-freshworks, line 4: other is not needed

hitankar-freshworks (changes requested):
added couple of suggestions

fwpushan:
> If the file main->config file change isn't needed lets revert it

@arranfw  There are multiple failure on the code de-coupling solve one of them.

vyasworks:
> > If the file main->config file change isn't needed lets revert it
> 
> @arranfw There are multiple failure on the code de-coupling solve one of them.

can you please list them and what each one solves?

apps/api/src/main.ts
```diff
@@ -1,92 +1,4 @@
...
-  });
-
-  return {
-    app,
-    expressApp,
-  };
-}
+import { createNestApp } from './app.config';
 
 async function bootstrap() {
```
fwpushan, line 3: @arranfw @hitankar-freshworks @dbayly-freshworks @vyasworks 
Decoupling bootstrap code from nest app config code **Fix multiple default connection issue**

apps/api/src/form/entity/form.entity.ts
```diff
@@ -6,7 +6,7 @@ import { FormPayloadDTO } from '@ehpr/common';
 @Entity('form')
 export class FormEntity extends BaseEntity {
   @Exclude()
-  @Column('jsonb', { nullable: false })
+  @Column('jsonb', { nullable: true })
```
fwpushan, line 9: @arranfw @hitankar-freshworks @vyasworks @dbayly-freshworks 
This is not required we can revert this.

apps/api/src/form/form.service.ts
```diff
@@ -24,15 +26,22 @@ export class FormService {
...
+    if (!newForm.version) {
+      newForm.version = '1.0.0';
+    }
     const savedForm = await this.formRepository.save(newForm);
-
-    const notifiedForm = await this.sendMail(savedForm);
-
-    return notifiedForm;
+    this.logger.log(`Saved form with id ${savedForm.id}`);
+    if (savedForm.payload?.contactInformation.email) {
```
fwpushan, line 34: @arranfw @hitankar-freshworks @dbayly-freshworks @vyasworks 
Any improper payload can break API, so this defensive code is introduce .... but no effect if validation enables in system

apps/api/src/mail/mail.service.ts
```diff
@@ -30,48 +29,43 @@ export class MailService {
...
-          auth: {
-            username: process.env.CHES_CLIENT_ID as string,
-            password: process.env.CHES_CLIENT_SECRET as string,
-          },
-          headers: {
-            'Content-Type': 'application/x-www-form-urlencoded',
-          },
+    const token = await axios.post(
+      process.env.CHES_AUTH_URL as string,
+      stringify({ grant_type: 'client_credentials' }),
```
fwpushan, line 56: @vyasworks @hitankar-freshworks @arranfw @dbayly-freshworks 
some unknown issue is occurred regarding usage of `URLSearchParams` to construct request body. May be that is not well supported by Node.

arranfw: This is surprising, URLSearchParams looks fully supported and has worked well in the past, maybe we can revisit this implementation before opting for a library over native functionality

apps/api/src/form/form.service.ts
```diff
@@ -24,15 +26,22 @@ export class FormService {
       confirmationId,
     } as Partial<FormEntity>);
 
+    if (!newForm.version) {
+      newForm.version = '1.0.0';
```
arranfw, line 30: Should we default assume un-specified submissions are 1.0.0? This also doesn't fit the format I've seen previously, I thought the value was `v1`

apps/api/src/form/form.service.ts
arranfw: I'd be comfortable with a TODO here, we don't want this staying past go-live

fwpushan:
@hitankar-freshworks @vyasworks @arranfw @dbayly-freshworks Please check my comments on code.

apps/api/src/form/form.service.ts
fwpushan: Again this is defensive code when validation is not present

apps/api/src/mail/mail.service.ts
fwpushan: `qs` is node js native lib

arranfw: Egg on my face!

arranfw (approved):

arranfw (approved):

PR #75, opened 2021-12-16
Use LZ2 & Terraform Cloud

vyasworks:


fwpushan (changes requested):
1. Rename existing dir, we don't want to configure vpc or other fw specific items
2. We need older IaC to tear down fw
3. your existing igonored state vars can create noise here.

fwpushan:
Work accomplished by PR-81, closing this

PR #76, opened 2021-12-16
EHPR-86 | Confirmation Page with ID

sunghwan-park-fw:


apps/web/src/pages/_app.tsx
```diff
@@ -6,15 +6,14 @@ import type { AppProps } from 'next/app';
 import { Footer, Header } from '@components';
 
 axios.defaults.baseURL = process.env.NEXT_PUBLIC_API_URL;
+console.log('backend url = ', axios.defaults.baseURL);
```
dbayly-freshworks, line 9: Stray console.log, might be good to remove it.

dbayly-freshworks (approved):
Looks mostly fine to me. One stray console.log

apps/web/src/pages/_app.tsx
sunghwan-park-fw: Removed.
I wonder why NEXT_PUBLIC_API_URL in .env.local is not applied after restarting by file changes. I need to restart the container.

apps/web/src/pages/_app.tsx
```diff
@@ -11,10 +11,8 @@ function MyApp({ Component, pageProps }: AppProps) {
   return (
     <div className='h-full flex flex-col'>
       <Header />
-      <main className='flex-grow bg-bcLightBackground flex justify-center md:pt-11 pt-5'>
-        <div className='h-min md:w-layout w-full md:mx-0 mx-2 mb-12'>
-          <Component {...pageProps} />
-        </div>
+      <main className='flex-grow flex justify-center'>
```
fwpushan, line 14: Any particular reason for this change?

arranfw: Designs have different background colors for the form and the confirmation screen so pages need to maintain their own layouts

apps/web/src/pages/_app.tsx
arranfw: .env files typically are only applied on application start aren't they?

apps/web/src/pages/_app.tsx
sunghwan-park-fw: Every form page's background is gray but confirmation's white.

apps/web/src/pages/confirmation.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ExternalLink, Notice, Button } from '@components';
+
+interface WithRouterProps {
+  router: NextRouter;
+}
+
+const Confirmation: React.FC<WithRouterProps> = props => {
+  const nextRouter = useRouter();
+  const { router } = props;
+  const id = router?.query?.id;
```
arranfw, line 12: I don't love that this page can be access by simply filling in an ID manually, this could arguably be better as react state but I don't think it's a merge blocker @fwpushan @dbayly-freshworks what do you think of storing the ID in the url?

arranfw: There might be a case for this being a sensitive code as it can be used to retrieve PII and storing it in a URL might not be the best option

sunghwan-park-fw: even though id is passed as a query param, it is null if the page is not redirected by router.

arranfw (approved):

PR #81, opened 2021-12-17
EHPR-113 | Migrating to LZ2

fwpushan:


vyasworks:
let's rename this folder back to `terraform` after tearing down the temp env. 

this way we don't have to have a big diff.

arranfw (commented):
One small question, other comment is not a blocker

vyasworks (approved):

terraform/cloudfront.tf
```diff
@@ -30,34 +30,27 @@ resource "aws_cloudfront_origin_access_identity" "app" {
...
-  bucket = "${local.app_name}-logs"
-  acl    = "log-delivery-write"
-}
-
 resource "aws_cloudfront_distribution" "app" {
   comment = local.app_name
 
   aliases = local.has_domain ? [var.domain] : []
 
-  logging_config {
```
vyasworks, line 57: do we need to remove this?

terraform/network.tf
```diff
@@ -1,85 +1,88 @@
-resource "aws_vpc" "vpc" {
-  cidr_block           = "10.0.0.0/16"
-  enable_dns_support   = true
-  enable_dns_hostnames = true
-  tags = {
-    environment = local.namespace
-    Name        = local.namespace
-  }
+// Not working at the moment
```
vyasworks, line 1: This was a while back - Is this still the same case? have we tried to make it work?

apps/api/tsconfig.build.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "./tsconfig.json",
-  "exclude": ["node_modules", "test", "dist", "**/*spec.ts", "./env.js", "types"]
+  "exclude": ["node_modules", "test", "dist", "**/*spec.ts", "./env.js", "types", "ormconfig.js"]
```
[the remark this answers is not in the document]
fwpushan: We are not using this right now. if we keep js file in the root of the app, then other part of the app can easily import that.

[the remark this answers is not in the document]
fwpushan: * keeping this as included will change the **dist** directory struct.

terraform/network.tf
fwpushan: We will work on that.

terraform/cloudfront.tf
fwpushan: We will try to enable this while working on logging ticket.

PR #83, opened 2021-12-20
EHPR-131 | Preferences remaining

arranfw:
https://freshworks.atlassian.net/browse/EHPR-131

```
deploymentDuration: "fourToEight",
hasImmunizationTraining: true,
placementOptions: ["covid19Support", "homeSupport"]
```

image
image


packages/common/src/interfaces/form-payload.ts
```diff
@@ -42,17 +42,18 @@ export enum EmploymentCircumstances {
...
 }
 
 export enum DeploymentDurations {
-  ZERO_TO_TWO = 1,
-  TWO_TO_FOUR,
-  FOUR_TO_EIGHT,
-  EIGHT_PLUS,
+  TWO_TO_FOUR_WEEKS = 'twoToFour',
+  FOUR_TO_EIGHT = 'fourToEight',
+  EIGHT_PLUS = 'eightPlus',
```
dbayly-freshworks, line 58: This makes more sense than my method, I just hope we don't end up having to implement sorting on this column.

packages/common/src/dto/availability-information.dto.ts
```diff
@@ -26,9 +33,17 @@ export class AvailabilityDTO {
   })
   deploymentLocations!: LhaId[];
 
-  placementPrefs!: PlacementPreferencesDTO;
+  @IsArray({ message: 'Placement options are required' })
+  @ArrayMinSize(1, { message: 'Placement options are required' })
+  @ArrayMaxSize(Object.values(PlacementOptions).length, {
+    message: 'Invalid placement options',
+  })
```
dbayly-freshworks, line 40: Does this currently validate the contents of the array?

fwpushan (approved):

packages/common/src/dto/availability-information.dto.ts
arranfw: No I don't think so! Good catch

PR #86, opened 2021-12-21
EHPR-149 | Additional comments for non-clinical

arranfw:
https://freshworks.atlassian.net/browse/EHPR-149

### DIFF
--- a/apps/api/src/common/helper/csv/submissionExport.ts
+++ b/apps/api/src/common/helper/csv/submissionExport.ts
@@ -34,7 +34,7 @@ export enum SubmissionExportColumnHeaders {
   currentEmployment = 'Current Employment Type',
   employmentCircumstance = 'Employment Circumstance',
 
-  nonClinicalJobTitle = 'Additional Comments',
+  nonClinicalJobTitle = 'Non Clinical Job Title',
 }
 
 export interface SubmissionExportColumns {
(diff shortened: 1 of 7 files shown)


apps/api/src/common/helper/csv/formExport.ts
```diff
@@ -34,7 +34,7 @@ export enum FormExportColumnHeaders {
   currentEmployment = 'Current Employment Type',
   employmentCircumstance = 'Employment Circumstance',
 
-  additionalComments = 'Additional Comments',
+  nonClinicalJobTitle = 'Additional Comments',
```
dbayly-freshworks, line 37: The ticket is a bit confusing, however given that the title of the field is still `Additional Comments`, I feel a bit weird changing the key.

arranfw: Yeah I was a bit conflicted, we should probably change the header now that the content of the field has changed right?

dbayly-freshworks: Yeah, I'd double check with Erica on that one. It's possible that the intent of the field has changed, designs just haven't been update yet. 
I have a suspicion that they're asking for a field with two purposes though.

fwpushan (approved):

fwpushan:
@arranfw Please fix conflict

PR #91, opened 2021-12-21
Environment based migrations directories

arranfw:


terraform/api.tf
```diff
@@ -16,7 +16,7 @@ resource "aws_lambda_function" "api" {
 
   environment {
     variables = {
-      NODE_ENV           = var.target_env
+      TARGET_ENV           = var.target_env
```
fwpushan, line 19: we should specify NODE_ENV here...

fwpushan (changes requested):

fwpushan (approved):

PR #93, opened 2021-12-22
EHPR-134 | extension field not validated

dbayly-freshworks:
https://freshworks.atlassian.net/browse/EHPR-134

Screen Shot 2021-12-21 at 4 38 14 PM

### DIFF
--- a/apps/api/src/app.config.ts
+++ b/apps/api/src/app.config.ts
@@ -1,7 +1,7 @@
 import { NestFactory } from '@nestjs/core';
 import * as winston from 'winston';
 import { WinstonModule, utilities as nestWinstonModuleUtilities } from 'nest-winston';
-import { Logger, ValidationPipe } from '@nestjs/common';
+import { Logger } from '@nestjs/common';
 import { ExpressAdapter, NestExpressApplication } from '@nestjs/platform-express';
 import express from 'express';
 
--- a/apps/web/src/components/submission/components/Contact.tsx
+++ b/apps/web/src/components/submission/components/Contact.tsx
@@ -4,7 +4,7 @@ import { FormStepProps } from '.';
 import { SubmissionType } from '../validation';
 
 export const Contact: React.FC<FormStepProps> = ({ formKey }) => {
-  const { values, setFieldValue } = useFormikContext<SubmissionType>();
+  const { values } = useFormikContext<SubmissionType>();
   const fieldNames = {
     primaryPhone: `${formKey}.primaryPhone`,
     primaryPhoneExt: `${formKey}.primaryPhoneExt`,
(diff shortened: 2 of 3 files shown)


apps/api/src/app.config.ts
```diff
@@ -1,7 +1,7 @@
 import { NestFactory } from '@nestjs/core';
 import * as winston from 'winston';
 import { WinstonModule, utilities as nestWinstonModuleUtilities } from 'nest-winston';
-import { Logger, ValidationPipe } from '@nestjs/common';
+import { Logger } from '@nestjs/common';
```
arranfw, line 4: Stray change here

packages/common/src/dto/contact-information.dto.ts
```diff
@@ -28,8 +29,11 @@ export class ContactInformationDTO {
 
   @IsString()
   @IsOptional()
-  @Length(0, 255, {
-    message: '255 character limit',
+  @Length(0, 3, {
```
arranfw, line 32: This field was free-text by design because extensions can be complex or have multiple steps
> dial 3, then 320

The validation described in the ticket should be disabling the field if phone number hasn't been entered

arranfw (changes requested):

packages/common/src/dto/contact-information.dto.ts
arranfw: Was this for EHPR-134?

dbayly-freshworks: Yes! Ugh, I mislabled the branch.

apps/api/src/app.config.ts
dbayly-freshworks: Validation pipe was an unused import. I figured I would remove it when I saw it.

arranfw: Oh sorry about that, I read this as an addition 👍

apps/web/src/components/submission/components/Contact.tsx
```diff
@@ -1,7 +1,10 @@
 import { Field, FormStepHeader } from '@components';
+import { useFormikContext } from 'formik';
 import { FormStepProps } from '.';
+import { SubmissionType } from '../validation';
 
 export const Contact: React.FC<FormStepProps> = ({ formKey }) => {
+  const { values, setFieldValue } = useFormikContext<SubmissionType>();
```
fwpushan, line 7: I think setFieldValue is not used here.

fwpushan (commented):
LGTM,  please check my comment

apps/web/src/components/submission/components/Contact.tsx
dbayly-freshworks: Yup, I will remove it 👍

arranfw (commented):
What happens if the user enters an extension and then clears the phone number? Can they submit an extension without a paired phone number?

dbayly-freshworks:
> What happens if the user enters an extension and then clears the phone number? Can they submit an extension without a paired phone number?

It seems like a niche case, but I can fix that.

arranfw (approved):

fwpushan (approved):

PR #94, opened 2021-12-22
EHPR-25 | Setup test env and deployment pipeline

fwpushan:
- Test Env infra code
- Cleanup 
- Pipeline

### DIFF
--- a/.github/workflows/deploy-test.yml
+++ b/.github/workflows/deploy-test.yml
@@ -1,4 +1,4 @@
-name: Deployment to aws dev environment
+name: Deployment to aws test environment
 on:
   push:
     branches:
--- a/.github/workflows/deploy-test.yml
+++ b/.github/workflows/deploy-test.yml
@@ -4,18 +4,18 @@ on:
     branches:
       - main
     tags:
-      - dev
+      - test
 env:
   PROJECT: ehpr
   NODE_ENV: production
-  ENV_NAME: dev
+  ENV_NAME: test
   TF_VERSION: 1.1.1
 jobs:
   deploy:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     environment:
-      name: dev
+      name: test
     env:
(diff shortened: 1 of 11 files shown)


terraform/api.tf
```diff
@@ -75,16 +74,15 @@ resource "aws_apigatewayv2_stage" "api" {
   name        = "$default"
   auto_deploy = true
 
-  access_log_settings {
+  /*access_log_settings {
```
arranfw, line 77: Let's just delete this if we aren't using it, we can always refer to history if we need it again

.github/workflows/deploy-test.yml
```diff
@@ -0,0 +1,67 @@
+name: Deployment to aws test environment
```
arranfw, line 1: Wondering if there's any way to dedupe this deployment code, can we extract the tag name somehow? Tracking changes between our deployment workflows will get tough as time goes on

arranfw (changes requested):
Happy to push the workflow conversation into the new year

I think we should remove unneeded code rather than comment it out

apps/api/src/app.controller.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('AppController', () => {
 
   describe('root', () => {
     it('should return "Hello World!"', () => {
```
arranfw, line 18: can we update the test name too

.github/workflows/deploy-test.yml
fwpushan: we have different environment setup for this pipelines...the branch deployment will cause issue ...

fwpushan: we can not have conditional code in `environment.name`

terraform/api.tf
fwpushan: Fixed

arranfw (approved):

dbayly-freshworks (approved):
I don't see any red flags.  👍

PR #95, opened 2021-12-22
EHPR - 141 Added dashes to displayed code

dbayly-freshworks:
Screen Shot 2021-12-22 at 12 16 38 PM

Note - We will need to strip out dashes if we need to accept a confirmation code from the client. 

I've also tacked on EHPR -145 - add dashes to email template, and added some helper methods.

### DIFF
--- a/apps/api/src/submission/submission.service.ts
+++ b/apps/api/src/submission/submission.service.ts
@@ -40,7 +40,7 @@ export class SubmissionService {
     const { email } = payload.contactInformation;
     const mailable = new ConfirmationMailable({ email } as Recipient, {
       firstName: (payload.personalInformation as PersonalInformationDTO).firstName,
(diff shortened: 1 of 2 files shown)


fwpushan (approved):
LGTM

apps/api/src/submission/submission.service.ts
```diff
@@ -97,4 +97,18 @@ export class SubmissionService {
...
+      id.substring(0, 3) +
+      '-' +
+      id.substring(3, 6) +
+      '-' +
+      id.substring(6, 9) +
+      '-' +
+      id.substring(9)
+    );
+  }
+  private convertDashedIdToId(dashedId: string) {
```
arranfw, line 111: Looks like good candidates for a function in the common package

arranfw (approved):

PR #119, opened 2022-01-04
Added notice

dbayly-freshworks:
Screen Shot 2022-01-04 at 10 25 17 AM

### DIFF
--- a/apps/web/src/components/Advisory.tsx
+++ b/apps/web/src/components/Advisory.tsx
@@ -1,0 +1,13 @@
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+export const Advisory: React.FC = ({ children }) => {
+  return (
+    <div className='bg-bcLightBlueBackground py-4  text-left flex items-center'>
+      <div className='px-5 text-bcBluePrimary'>
+        <FontAwesomeIcon className='fa-lg' icon={faExclamationCircle}></FontAwesomeIcon>
+      </div>
+      <div className='text-bcBluePrimary'>{children}</div>
+    </div>
+  );
+};
--- a/apps/web/src/components/submission/components/Review.tsx
+++ b/apps/web/src/components/submission/components/Review.tsx
@@ -1,5 +1,4 @@
 import { Checkbox, Error, FormStepHeader, Link, Notice } from '@components';
-import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
 import {
   booleanToYesNo,
   EmploymentTypes,
@@ -24,7 +23,6 @@ import {
   deploymentDurationOptions,
   placementOptions as allPlacementOptions,
 } from '../validation/preferences';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Advisory } from 'src/components/Advisory';
 
 export const Review: React.FC = () => {
--- a/apps/web/src/components/submission/components/Review.tsx
+++ b/apps/web/src/components/submission/components/Review.tsx
@@ -1,4 +1,4 @@
-import { Checkbox, Error, FormStepHeader, Link, Notice } from '@components';
+import { Checkbox, Error, FormStepHeader, Link, Notice, Alert } from '@components';
 import {
   booleanToYesNo,
   EmploymentTypes,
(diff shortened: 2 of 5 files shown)


apps/web/src/components/submission/components/Review.tsx
```diff
@@ -154,6 +157,14 @@ export const Review: React.FC = () => {
           </Notice>
           <Error name='confirm' />
         </div>
+        <div className='bg-bcLightBlueBackground border-l-1 border-bcBluePrimary p-6 pb-5 md:pb-6 mb-5 text-left flex items-center'>
+          <div className='px-7 text-bcBluePrimary'>
+            <FontAwesomeIcon icon={faExclamationCircle}></FontAwesomeIcon>
+          </div>
+          <div className='text-bcBluePrimary'>
+            Please print and save this page for your own recrods.
```
arranfw, line 165:
```suggestion
            Please print and save this page for your own records.
```

arranfw (changes requested):
We should be able to match designs a bit closer with some changed margins, border radius and icon size. I think it could be worthwhile to abstract this into an alert component (with a single variant for now)

Placing the alert in the same div as the checkbox should help with spacing and will prevent the extra divider from showing

apps/web/src/components/Advisory.tsx
```diff
@@ -0,0 +1,13 @@
+import { faExclamationCircle } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+export const Advisory: React.FC = ({ children }) => {
```
arranfw, line 4: Can we call this `Alert` to match the BC gov term? https://developer.gov.bc.ca/Design-System/Alert-Banners

arranfw (changes requested):

apps/web/src/components/Advisory.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+    <div className='bg-bcLightBlueBackground py-4  text-left flex items-center'>
```
arranfw, line 6: the `rounded` class should be added here

apps/web/src/components/submission/components/Review.tsx
```diff
@@ -154,6 +156,7 @@ export const Review: React.FC = () => {
           </Notice>
           <Error name='confirm' />
         </div>
+        <Advisory>Please print and save this page for your own records.</Advisory>
```
arranfw, line 159: Moving this up one line to be after the error will help the layout

apps/web/src/components/Advisory.tsx
dbayly-freshworks: I thought about that, but the design doesn't have rounded corners.

arranfw: I see rounded on the version of designs I have 🤔  
image
https://www.figma.com/file/Cf5M3yWTLdJxngU6l6o17c/EHPR---Mockups?node-id=13%3A1211

apps/web/src/components/Advisory.tsx
dbayly-freshworks: Sure, but I feel weird about it for a couple reasons. 
1. Alert is already an html tag name, and does something very different 
2. The BC gov alert component has a lot more features/functions that what this component does. It feels like either a mismatched component or a fair bit of scope creep.

apps/web/src/components/Advisory.tsx
dbayly-freshworks: I see what happened, I had the component selected in figma, so it looked like it had short corners. It does feel a bit weird that this has rounded corners but the component above as sharp corners. 🤷 
 
Screen Shot 2022-01-04 at 10 20 47 AM

apps/web/src/components/Advisory.tsx
arranfw: It's just different variants and accessibility features isn't it? I think we should implement as much as is reasonable, accessibility stuff is pretty important but we don't need the other variants for now (warning/error/etc..)

apps/web/src/components/submission/components/Review.tsx
```diff
@@ -22,6 +23,7 @@ import {
   deploymentDurationOptions,
   placementOptions as allPlacementOptions,
 } from '../validation/preferences';
+import { Alert } from '../../Alert';
```
arranfw, line 26: Last small comment, could you add alert to `components/index.ts` and import as `from '@components'`

arranfw (approved):
🥳

PR #123, opened 2022-01-04
EHPR-184 | Logging API Validations

arranfw:
- Fix error filter logging errors twice
- Fix error filter private keys implementation
- Fix LHA array validator naming
- Filter payload from error so we don't log PII
- Fix error logging parameters
- Use `defaultMessage` + `exceptionFactory` to expose validation error messages

image

I really hamfisted through this so I'm expecting some comments on how to clean this up a bit :\ also the response is heavily nested arrays which I think checks out because of the nested nature of our data, but maybe we can improve that as well


dbayly-freshworks (approved):
Looks good to me 👍

PR #124, opened 2022-01-05
EHPR-117 | Slack Alerting

arranfw:
- Winston error logging sends to Slack
- Sample error endpoint
- Tested console + slack for regular, HTTP, axios, and validation exceptions
- Improved some error handling
- Error logger assumes you're passing an error object

image
image
image- 
image
image

image


.github/workflows/deploy-dev.yml
```diff
@@ -12,7 +12,7 @@ env:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
```
fwpushan, line 15: please rebase

apps/api/src/common/slack.ts
```diff
@@ -0,0 +1,35 @@
+import { Logger } from '@nestjs/common';
+import axios from 'axios';
+import yaml from 'js-yaml';
+import https from 'https';
+
+const webhookUrl = process.env.SLACK_ALERTS_WEBHOOK_URL;
+
+export default function postToSlack(data: unknown): void {
```
fwpushan, line 8: we should handle slack push failure gracefully. This is not main ops.... so in any case it should not tigger any failure. Mean, put whole code into a try/catch  otherwise that can create recursive execution.

apps/api/src/app.controller.ts
```diff
@@ -1,12 +1,26 @@
...
   getVersion(): object {
     return this.appService.getVersionInfo();
   }
+
+  @ApiOperation({
+    summary: 'Trigger a sample error',
+  })
+  @ApiResponse({ status: HttpStatus.INTERNAL_SERVER_ERROR })
+  @Get('/error')
+  getError(): object {
```
fwpushan, line 23: should we keep this as part of main app?

apps/api/src/common/logger.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (axios.isAxiosError(e)) {
+      message = (e as AxiosError).response?.data;
+    }
+
+    // For handling manually crafted validation error message arrays, see 'exceptionFactory' in 'app.config.ts'
+    if (error.response?.message) {
+      message = error.response?.message;
+    }
+
+    postToSlack({ message, stack: error.stack, context });
```
fwpushan, line 44: Make sure this should not invoke recursive execution.

apps/api/src/mail/mail.service.ts
```diff
@@ -1,16 +1,17 @@
...
 import * as handlebars from 'handlebars';
+
 import { Mailable } from './mailables/mail-base.mailable';
 import { MailOptions } from './mail-options.interface';
 import { ChesResponse } from './types/ches-response';
 
 @Injectable()
 export class MailService {
-  constructor(@Inject(Logger) private readonly logger: LoggerService) {
+  constructor() {
```
fwpushan, line 14: Do our application error filter provides enough context to debug different failure stages of this service?

fwpushan (changes requested):
Please check my comments.

dbayly-freshworks (approved):

apps/api/src/mail/mail.service.ts
arranfw: The original error will be caught in the controller and logged, so we'll know what the failed response was as well as what the original error is

apps/api/src/common/logger.service.ts
```diff
@@ -30,10 +30,19 @@ export class AppLogger implements LoggerService {
...
+      message = e;
+    }
 
     if (axios.isAxiosError(e)) {
-      message = (e as AxiosError).response?.data;
+      const { response, config } = e;
+      message = {
+        url: config.url,
+        method: config.method,
+        ...(response?.data ? { data: response.data } : {}),
```
fwpushan, line 44: what if data contains any sensitive info or PI?

fwpushan (approved):
LGTM just one comment

apps/api/src/common/logger.service.ts
arranfw: This response would be what the external API is returning as error data also in this implementation the logger isn't responsible for filtering

We've added a bit of filtering in the error filter to prevent logging of the submission payload
We should also be diligent in not explicitly logging PII through info/warn methods

PR #131, opened 2022-01-07
EHPR-195 Added before insert function to delete unvalidated fields

dbayly-freshworks:


apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -19,4 +19,41 @@ export class SubmissionEntity extends BaseEntity {
   @Exclude()
   @Column('varchar', { nullable: false })
   version!: string;
+  @BeforeInsert()
+  beforeInsert() {
+    let { contactInformation, skillInformation, availabilityInformation } = this.payload;
+
+    // Remove specialties if the stream is non clinical
+    if (skillInformation.stream && skillInformation.stream !== streamsById.Nonclinical.id) {
```
arranfw, line 27: I think we'll need to also check for streams like midwife that shouldn't have specialties, something like a new helper in the common package `getStreamsWithoutSpecialties` could be use

apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -19,4 +19,41 @@ export class SubmissionEntity extends BaseEntity {
...
+      this.payload.skillInformation.specialties = [];
+    }
+
+    // Remove Health authorities if the user is not a resident or employed
+    if (
+      [EmploymentTypes.HEALTH_SECTOR_EMPLOYED, EmploymentTypes.HEALTH_SECTORY_RESIDENCY].includes(
```
arranfw, line 33: Would it be helpful to abstract this conditions and reuse them in API and Common?

apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -19,4 +19,41 @@ export class SubmissionEntity extends BaseEntity {
...
+      delete this.payload.skillInformation.nonClinicalJobTitle;
+    }
+
+    // Remove deployment locations if the user agrees to deploy anywhere.
+    if (availabilityInformation.deployAnywhere) {
+      this.payload.availabilityInformation.deploymentLocations = [];
+    }
+
+    // Remove secondary phone if there is no primary
+    if (!contactInformation.primaryPhone) {
```
arranfw, line 54: `primaryPhone` is required so this might not be necessary

apps/api/src/submission/entity/submission.entity.ts
arranfw: Actually this is captured by [EHPR-177](https://freshworks.atlassian.net/browse/EHPR-177), would be nice to add to reduce possible confusion in the future, what do you think?

arranfw (changes requested):
I'm not sure what the best default values are for some of these, undefined/null/[]/'' 🤷 I think it's a bit inconsistent right now

dbayly-freshworks:
> Good investigation 👍 this should work
> 
> I'm not sure what the best default values are for some of these, undefined/null/[]/'' 🤷 I think it's a bit inconsistent right now

i think the pattern I will follow is that arrays will get an empty array, and all others will get undefined. 
There are still a few issues to sort out, mainly to do with the fields being required.

fwpushan (commented):
Just address the comment from @arranfw otherwise looks ok to me

apps/api/src/submission/entity/submission.entity.ts
dbayly-freshworks: I don't think it's quite needed. If we run into this again, then it might make more sense.

apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -21,16 +22,19 @@ export class SubmissionEntity extends BaseEntity {
...
   @BeforeInsert()
   beforeInsert() {
-    let { contactInformation, skillInformation, availabilityInformation } = this.payload;
+    let { skillInformation, availabilityInformation } = this.payload;
 
     // Remove specialties if the stream is non clinical
-    if (skillInformation.stream && skillInformation.stream !== streamsById.Nonclinical.id) {
+    if (
+      skillInformation.stream !== streamsById.Nonclinical.id ||
+      getStreamById(skillInformation.stream).specialties.length === 0
```
arranfw, line 30: 💯

apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -1,7 +1,8 @@
 import { Entity, Column, BeforeInsert } from 'typeorm';
 import { Exclude } from 'class-transformer';
 import { BaseEntity } from 'src/database/base.entity';
-import { EmploymentTypes, streamsById, SubmissionPayloadDTO } from '@ehpr/common';
+import { EmploymentTypes, getStreamById, streamsById, SubmissionPayloadDTO } from '@ehpr/common';
+import { stream } from 'winston';
```
arranfw, line 5: Accidental import?

apps/api/src/submission/entity/submission.entity.ts
```diff
@@ -19,4 +20,38 @@ export class SubmissionEntity extends BaseEntity {
   @Exclude()
   @Column('varchar', { nullable: false })
   version!: string;
+  @BeforeInsert()
+  beforeInsert() {
+    let { skillInformation, availabilityInformation } = this.payload;
+
+    // Remove specialties if the stream is non clinical
```
arranfw, line 27:
```suggestion
    // Remove specialties if the stream is non clinical or if selected stream has no specialties
```

packages/common/src/dto/skill-information.dto.ts
```diff
@@ -73,7 +73,7 @@ export class SkillInformationDTO {
   @ValidateIf(o => EmploymentTypes.NOT_HEALTH_SECTOR_EMPLOYED === o.currentEmployment)
   @IsIn(Object.values(EmploymentCircumstances), { message: 'Invalid circumstance selection' })
   @IsString({ message: 'Circumstance selection is required' })
-  employmentCircumstance!: EmploymentCircumstances;
+  employmentCircumstance!: EmploymentCircumstances | undefined;
```
arranfw, line 76:
```suggestion
  employmentCircumstance?: EmploymentCircumstances;
```

Does this make more sense? or is there a difference

arranfw (commented):
Couple small comments! Looks great

apps/api/src/submission/entity/submission.entity.ts
dbayly-freshworks: Yup 😞

packages/common/src/dto/skill-information.dto.ts
dbayly-freshworks: At this point I think it's equivalent.

arranfw: Seems weird to declare it non-null and then say it can be undefined :P

arranfw (approved):

PR #133, opened 2022-01-07
Prod Bastion Host

fwpushan:
- Add easy env switching for db tunnel
- Add some info logs in db tunnel


arranfw (approved):
Should we add a short tutorial in the readme for this?

This requires setting `ENV_NAME` in the terminal to use test/prod right?

fwpushan:
> Should we add a short tutorial in the readme for this?
> 
> This requires setting `ENV_NAME` in the terminal to use test/prod right?

I'm setting value in env file right now ....

PR #134, opened 2022-01-07
Optimize FE loadtimes + UptimeRobot debugging

arranfw:
- We weren't purging CSS which had some really bad performance/load time consequences.
- Keep the version endpoint broken for now until we can debug UptimeRobot

image
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,6 +169,7 @@ build-api:
 
 build-web:
 	@echo "++\n***** Building Web for AWS\n++"
+	@NODE_ENV=production && yarn workspace @ehpr/web build
 	@NODE_ENV=production && yarn workspace @ehpr/web export
 	@mv ./apps/web/out ./terraform/build/app
 	@echo "++\n*****"
--- a/Makefile
+++ b/Makefile
@@ -169,8 +169,7 @@ build-api:
 
 build-web:
 	@echo "++\n***** Building Web for AWS\n++"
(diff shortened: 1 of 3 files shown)


apps/web/tailwind.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  purge: ['./pages/**/*.tsx', './components/**/*.tsx'],
+  purge: ['./src/pages/**/*.tsx', './src/components/**/*.tsx'],
```
dbayly-freshworks, line 2: Good catch!

dbayly-freshworks (approved):

Makefile
```diff
@@ -169,8 +169,8 @@ build-api:
 
 build-web:
 	@echo "++\n***** Building Web for AWS\n++"
-	@yarn workspace @ehpr/web build
-	@yarn workspace @ehpr/web export
+	@NODE_ENV=production && yarn workspace @ehpr/web build
+	@NODE_ENV=production && yarn workspace @ehpr/web export
```
dbayly-freshworks, line 173: This feels a bit sus though. Would you be able to elaborate a bit on what this is doing?

arranfw: These commands should set `NODE_ENV=production` but we've manually set it to `development` at the start of the makefile so it's using that value unless we manually override it again.

We need `NODE_ENV=production` to signal to purge unused CSS among other likely optimizations

PR #135, opened 2022-01-07
Remove NODE_ENV=development

arranfw:
I think `NODE_ENV=development` is an anti-pattern and it's adding a bit of confusion, let's discuss these changes and merge if it's an improvement


fwpushan (approved):
LGTM. But make sure next build and import generating expected outcome.

PR #139, opened 2022-01-10
Added zap scan to PR check to test it out

dbayly-freshworks:


arranfw:
What's the procedure/plan to notify about these issues? Is the only way for now to navigate to the action and download the ZIP?

dbayly-freshworks:
> What's the procedure/plan to notify about these issues? Is the only way for now to navigate to the action and download the ZIP?

That looks to be the plan. I could remove the -I flag (it was enabled on HCAP) to have the action fail, which would alert the dev running the action. What do you think?

.github/workflows/deploy-dev.yml
```diff
@@ -65,3 +65,9 @@ jobs:
           make plan
           make deploy-infra
           make deploy-app
+
```
fwpushan, line 68: Do we need to wait a bit to ensure deployment ?

arranfw: I believe our TF process waits for cloudfront to be fully updated before it finishes

arranfw (approved):

PR #143, opened 2022-01-11
Ehpr 112 cypress testing

dbayly-freshworks:
Added local cypress testing and testing for the first three pages. 

All the cydata props are to allow cypress to accurately select elements for testing.


apps/web/cypress/integration/partone.spec.js
```diff
@@ -0,0 +1,16 @@
+describe('load page.com', () => {
```
arranfw, line 1: Let's update these describe blocks

apps/web/src/components/form/Field.tsx
```diff
@@ -37,6 +39,7 @@ export const Field: React.FC<FieldProps> = props => {
       <FormikField
         id={name}
         aria-describedby={description ? `${name}-description` : null}
+        data-cy={`${cyData}`}
```
arranfw, line 42: Is a template needed here?
```suggestion
        data-cy={cyData}
```

apps/web/src/components/form/MultiSelect.tsx
```diff
@@ -12,12 +12,13 @@ interface MultiSelectProps extends FieldProps {
...
  */
 export const MultiSelect: React.FC<MultiSelectProps> = props => {
-  const { label, name, options, disabled } = props;
+  const { label, name, options, disabled, cyData } = props;
 
   return (
     <Field
       name={name}
       label={label}
+      cyData={cyData}
```
arranfw, line 21: If we rename this `data-test-id` it'll make more sense to reuse the test id for unit tests as well

apps/web/tsconfig.json
```diff
@@ -21,6 +21,11 @@
       "@assets/*": ["public/assets/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": [
+    "next-env.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    "cypress/integration/1-getting-started/partone.spec.js"
```
arranfw, line 28: Is this complete?

apps/web/cypress/integration/partthree.spec.js
```diff
@@ -0,0 +1,26 @@
+describe('load page.com', () => {
+  beforeEach(() => {
+    cy.visit('/submission/1');
+    // Fill out first page
+    cy.get(`[data-cy=firstname]`).type('testname').should('have.value', 'testname');
```
arranfw, line 5: This duplication is weird, when we change a field we need to update multiple tests right? Not sure what the best implementation here is but I think if we end up needing all of this duplication we should just make it a single test

arranfw (changes requested):
Can we name the cypress tests based on their UX names?

Primary
Contact
Credentials
Preferences
Review

Also what do you think of changing `cyData` to `data-test-id` to be more reusable by other test suites?

apps/web/tsconfig.json
dbayly-freshworks: I think this is auto generated, I'll clear it out.

apps/web/src/components/form/Field.tsx
dbayly-freshworks: I don't think so. cyData should always be a string

apps/web/cypress/integration/partthree.spec.js
dbayly-freshworks: It's a bit jank, but I don't see how to access pages further in the process when we have a redirect in effect. It may end up that we have to have this all in one very large final test.

dbayly-freshworks: I'll definitely remove all the 'should' calls here. We don't need to test anything in the before step

arranfw: I've disabled the form redirection when `NODE_ENV === test` in the past, but I think that's an antipattern. I think I'd like to see this as a single test

dbayly-freshworks:
> Can we name the cypress tests based on their UX names?
> 
> Primary Contact Credentials Preferences Review
> 
> Also what do you think of changing `cyData` to `data-test-id` to be more reusable by other test suites?

Absolutely for the first part. 

The second part I'm a bit fuzzier on. I think a deeper discussion on testing strategy may be needed if this project continues. 

I feel I owe a bit of an explanation. This PR is pretty rough because cypress testing was descoped. This is mostly WIP. It may be worth just keeping it on the branch, but then it's pretty likely that it'll be forgotten.

arranfw:
Let's mark this as a draft/close it and put the ticket back in Todo in that case

PR #144, opened 2022-01-12
EHPR-202 | Export Script + Merge Scripts Package

arranfw:
https://freshworks.atlassian.net/browse/EHPR-202


fwpushan (commented):
LGTM to me but I saw some scripts/out element. Can you check and make sure they are not ignored.

packages/scripts/data/locations.csv
```diff
@@ -0,0 +1,85 @@
+Health Authority,Health Service Delivery Areas,Local Health Areas
```
dbayly-freshworks, line 1: Since the source of truth for will be an extract, maybe it would be good to gitignore both csvs

packages/scripts/README.md
```diff
@@ -0,0 +1 @@
+# scripts
```
dbayly-freshworks, line 1: Nit: it could be useful to add some documentation here.

.gitignore
```diff
@@ -18,3 +18,5 @@ node_modules
 .build
 yarn.build.json
 ssh-keypair*
+
+scripts/out
```
dbayly-freshworks, line 22: Might be good to delete the git keep file for this

dbayly-freshworks (approved):
It's a bit hard to judge, but I think the risks for this script is low. If these csv extracts do end up having issues we can tweak the script as we use it. 

I have a few nits, but overall I think this is good.

.gitignore
arranfw: Did I mess this up? I wanted to keep the directory but ignore it's contents

arranfw:
Fixed the in/out folders, the intent was to keep the folders but ignore the children

dbayly-freshworks (approved):
Looks good

fwpushan (approved):

PR #145, opened 2022-01-12
EHPR-203 Fixed extra large icons

dbayly-freshworks:
Fixed Extra large icons issue 
Fixed issue whereby subspecialties were getting stripped out of the entity on submission
Screen Shot 2022-01-12 at 12 49 31 PM
Screen Shot 2022-01-12 at 12 49 23 PM

### DIFF
--- a/apps/api/src/mail/mail.service.ts
+++ b/apps/api/src/mail/mail.service.ts
@@ -77,7 +77,7 @@ export class MailService {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async sendMailable(mailable: Mailable<any>): Promise<ChesResponse> {
     const mailOptions: Partial<MailOptions> = {
-      from: process.env.MAIL_FROM || 'EHPRDoNotReply@gov.bc.ca',
+      from: process.env.MAIL_FROM,
       to: [mailable.recipient.email],
       subject: mailable.subject,
     };
--- a/apps/api/src/mail/mail.service.ts
+++ b/apps/api/src/mail/mail.service.ts
@@ -77,7 +77,7 @@ export class MailService {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async sendMailable(mailable: Mailable<any>): Promise<ChesResponse> {
     const mailOptions: Partial<MailOptions> = {
-      from: process.env.MAIL_FROM,
+      from: process.env.MAIL_FROM || 'EHPRDoNotReply@gov.bc.ca',
       to: [mailable.recipient.email],
       subject: mailable.subject,
     };
--- a/apps/web/src/components/Alert.tsx
+++ b/apps/web/src/components/Alert.tsx
@@ -4,7 +4,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 export const Alert: React.FC = ({ children }) => {
   return (
     <div className='bg-bcLightBlueBackground py-4  text-left flex items-center rounded'>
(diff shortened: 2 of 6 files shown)


apps/api/src/mail/mail.service.ts
```diff
@@ -77,7 +77,7 @@ export class MailService {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async sendMailable(mailable: Mailable<any>): Promise<ChesResponse> {
     const mailOptions: Partial<MailOptions> = {
-      from: process.env.MAIL_FROM,
+      from: process.env.MAIL_FROM || 'EHPRDoNotReply@gov.bc.ca',
```
arranfw, line 80: Why do we need this change?

apps/web/src/components/Alert.tsx
```diff
@@ -4,7 +4,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 export const Alert: React.FC = ({ children }) => {
   return (
     <div className='bg-bcLightBlueBackground py-4  text-left flex items-center rounded'>
-      <div className='px-5 text-bcBluePrimary'>
+      <div className='px-5 text-bcBluePrimary' style={{ width: '60px' }}>
```
arranfw, line 7: 🙅  `h-5` or `w-5` on the icon instead of `fa-lg` should work here

apps/web/src/components/Disclosure.tsx
```diff
@@ -15,10 +15,12 @@ export const Disclosure: React.FC<DisclosureProps> = ({ buttonText, content }) =
...
           <HeadlessDisclosure.Button className={'flex justify-between items-center w-full'}>
             {buttonText}
-            <FontAwesomeIcon
-              icon={faChevronDown}
-              className={classnames('text-gray-500 mr-5', { 'transform rotate-180': open })}
-            />
+            <div style={{ width: '50px' }}>
+              <FontAwesomeIcon
+                icon={faChevronDown}
+                className={classnames('text-gray-500 mr-5', { 'transform rotate-180': open })}
```
arranfw, line 21: Same thing, we should almost never need inline styles 
```suggestion
                className={classnames('text-gray-500 mr-5 h-5', { 'transform rotate-180': open })}
```

arranfw (changes requested):
This would be nice as 3 separate PRs and I think we need a bug ticket for the specialties bug

RC for inline styles + email bit

apps/web/src/components/Disclosure.tsx
dbayly-freshworks: Something funky is going on with tailwind. Your suggestion works, but when I translate it to the div above, it doesn't work. It seems 'h-5' != width:'50px'. 

Thats why I went with style, because it didn't work initially with tailwind. 
Screen Shot 2022-01-12 at 1 03 58 PM

arranfw: Looks like it's because the style is applied to a wrapper div instead of the icon itself

arranfw: h-5 is 1.25rem not 50px

apps/api/src/mail/mail.service.ts
dbayly-freshworks: I'll remove it. I didn't have the environment variable in my local, so CHES was complaining. Also, it doesn't hurt to have a backup just in case the variable isn't defined in an environment

arranfw (approved):

PR #147, opened 2022-01-12
EHPR-205 Added previous deployment question

dbayly-freshworks:
Screen Shot 2022-01-12 at 3 46 10 PM

### DIFF
--- a/apps/api/test/fixture/validSubmission.json
+++ b/apps/api/test/fixture/validSubmission.json
@@ -43,7 +43,8 @@
       "placementOptions": ["other", "anywhere"],
       "hasImmunizationTraining": true,
       "deploymentDuration": "twoToFour",
-      "deploymentType": ["fullTime"]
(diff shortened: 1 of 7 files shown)


apps/api/test/fixture/validSubmission.json
```diff
@@ -42,7 +42,8 @@
       ],
       "placementOptions": ["other", "anywhere"],
       "hasImmunizationTraining": true,
-      "deploymentDuration": "twoToFour"
+      "deploymentDuration": "twoToFour",
+      "hasPreviousDeployment": "yes"
```
arranfw, line 46: I think we should add this to the invalidSubmission fixture as well to keep it as the single error for firstName?

arranfw (commented):
I added a commit to fix the error not showing up when submitting the page without answering the question, I left a small question but otherwise this looks good

PR #148, opened 2022-01-13
EHPR-204 Copy and options changes requested in demo

dbayly-freshworks:
Changed Vancouver Region to Vancouver Costal 
Added outside BC option for employment location 
Screen Shot 2022-01-12 at 4 37 55 PM
Screen Shot 2022-01-12 at 4 37 45 PM

### DIFF
--- a/apps/web/src/components/submission/components/Preferences.tsx
+++ b/apps/web/src/components/submission/components/Preferences.tsx
@@ -121,7 +121,7 @@ const DeploymentLocationSelector: React.FC = () => {
               <span className='font-bold text-black p-5'>
                 {label} (
                 <a
-                  href={HaPdfSizeMap[value]?.url}
+                  href={HaPdfSizeMap[value].url}
                   target='_blank'
                   rel='noreferrer'
                   className='font-bold text-bcBlueLink'


packages/common/src/data/location_data.json
```diff
@@ -1,9 +1,9 @@
 {
   "has": {
     "byId": {
-      "VancouverRegion": {
-        "id": "VancouverRegion",
-        "name": "Vancouver Region",
+      "VancouverCostal": {
```
arranfw, line 4: Can we make sure to update the source data for this and rerun the script to verify the output?

packages/scripts/src/submission_export.js
```diff
@@ -33,7 +33,7 @@ export const formExportColumnHeaders = [
 
   // Preferences
   { id: 'deployAnywhere', title: 'Deploy Anywhere' },
-  { id: 'VancouverRegion', title: 'Vancouver Region' },
+  { id: 'VancouverCostal', title: 'Vancouver/Coastal' },
```
arranfw, line 36: Something to think about is how we'll handle changing copy in the future without changing the IDs, no biggie right now because we're pre-live

packages/common/src/data/location_data.json
```diff
@@ -1,9 +1,9 @@
...
+        "id": "VancouverCostal",
```
arranfw, line 5: Looks like a typo in a few places
```suggestion
        "id": "VancouverCoastal",
```

arranfw (changes requested):
RC for updating the locations data source and typo

packages/common/src/data/location_data.json
arranfw, line 4: One last costal here

PR #155, opened 2022-03-09
Fix specialty selection

sunghwan-park-fw:
- show disabled select options
- remove invalid subspecialties when changing a specialty


dbayly-freshworks (approved):
Looks ok to me

PR #159, opened 2022-03-10
nextjs secure header

sunghwan-park-fw:


apps/web/next.config.js
```diff
@@ -14,4 +17,30 @@ module.exports = {
       },
     ];
   },
+  poweredByHeader: false,
+  async headers() {
```
vyasworks, line 21: We are using static hosting in cloudfront. Have you  tried deploying to dev to see if this works? 

@dbayly-freshworks is working on the same area too. let's sync so there is no repetition of work.

sunghwan-park-fw: I guessed this would work only when we use next.js to serve.
In lamda, we need to update response.js, right? 
It seems David is working on it so I just wanted to verify the headers locally.

vyasworks: Yes, however, this would not be useful in our AWS environments. 

you have to make changes here https://github.com/bcgov/ehpr2/blob/980c4834f4adfbb795d489100a67d533a3e4782e/terraform/cloudfront/response.js#L1

dbayly-freshworks (approved):

PR #161, opened 2022-03-11
Set server as '*" and add cache control header

vyasworks:
Also https://github.com/bcgov/ehpr2/issues/4

### DIFF
--- a/terraform/cloudfront/response.js
+++ b/terraform/cloudfront/response.js
@@ -14,7 +14,7 @@ function handler(event) {
   headers['x-xss-protection'] = { value: '1; mode=block' };
   headers['x-download-options'] = { value: 'noopen' };
   headers['server'] = { value: '*' };
-  headers['cache-control'] = { value: 'no-cache, no-store, must-revalidate' };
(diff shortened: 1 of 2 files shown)


terraform/cloudfront/response.js
```diff
@@ -13,7 +13,8 @@ function handler(event) {
   headers['x-frame-options'] = { value: 'DENY' };
   headers['x-xss-protection'] = { value: '1; mode=block' };
   headers['x-download-options'] = { value: 'noopen' };
-
+  headers['server'] = { value: '*' };
+  headers['cache-control'] = { value: 'no-cache, no-store, must-revalidate' };
```
sunghwan-park-fw, line 17: Mozilla says that `no-store` is most restrictive and other directives are ineffective.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
```
# conflicted
Cache-Control: private, no-cache , no-store, max-age=0, must-revalidate

# equivalent to
Cache-Control: no-store
```

sunghwan-park-fw (approved):

dbayly-freshworks (approved):

PR #165, opened 2022-03-16
EHPR-230 Added delete functionality to each specialty row

dbayly-freshworks:
See post on slack for video on functionality

### DIFF
--- a/apps/web/src/components/submission/components/Credential.tsx
+++ b/apps/web/src/components/submission/components/Credential.tsx
@@ -124,6 +124,8 @@ export const Credential: React.FC = () => {
                       index={index}
                       specialties={specialtyOptions}
                       subspecialties={subspecialties?.[index]}
+                      deleteFunction={() => arrayHelpers.remove(index)}
+                      enableDelete={specialties.length > 1}
                     />
                   ))}
                 </fieldset>
@@ -141,21 +143,6 @@ export const Credential: React.FC = () => {
                       onClick={() => arrayHelpers.push({ ...defaultSpecialtyValue })}
                     >
                       Add
-                    </button>
-                  ) : null}
-                  {specialties.length > 1 && specialties.length !== specialtyOptions?.length ? (
-                    <span aria-hidden className='text-sm mx-1'>
-                      |
-                    </span>
-                  ) : null}
-                  {specialties.length > 1 ? (
-                    <button
-                      type='button'
-                      className='text-bcRedError'
-                      aria-label='delete the last specialty'
-                      onClick={() => arrayHelpers.pop()}
-                    >
-                      Delete
                     </button>
                   ) : null}
                 </div>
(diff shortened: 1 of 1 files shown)


apps/web/src/components/submission/components/Credential.tsx
```diff
@@ -124,6 +124,8 @@ export const Credential: React.FC = () => {
                       index={index}
                       specialties={specialtyOptions}
                       subspecialties={subspecialties?.[index]}
+                      deleteFunction={() => arrayHelpers.remove(index)}
+                      enableDelete={specialties.length > 1}
```
sunghwan-park-fw, line 128: Why do we need this?

apps/web/src/components/submission/components/Credential.tsx
```diff
@@ -221,6 +212,16 @@ const SpecialtySelector: React.FC<SpecialtySelectorProps> = ({
             />
           ))}
         </Select>
+        {enableDelete ? (
+          <button
+            type='button'
+            className='text-bcRedError'
+            aria-label='delete the last specialty'
```
sunghwan-park-fw, line 219: Now should it be n-th specialty?

apps/web/src/components/submission/components/Credential.tsx
```diff
@@ -221,6 +212,16 @@ const SpecialtySelector: React.FC<SpecialtySelectorProps> = ({
             />
           ))}
         </Select>
+        {enableDelete ? (
```
sunghwan-park-fw, line 215: if we want disallow deleting the last specialty, can we just check if index is 0?

apps/web/src/components/submission/components/Credential.tsx
dbayly-freshworks: The delete function handles the removal of the specialty.
The enable delete determines whether of not the delete button is shown. 

I put them at a higher level and pass them down into the component to keep the `SpecialtySelector` component generic and because otherwise I would need to pass the `specialties` array and `arrayHelper` function into the component.

dbayly-freshworks: The delete function handles the removal of the specialty.
The enable delete determines whether of not the delete button is shown. 

I put them at a higher level and pass them down into the component to keep the `SpecialtySelector` component generic and because otherwise I would need to pass the `specialties` array and `arrayHelper` function into the component.

apps/web/src/components/submission/components/Credential.tsx
dbayly-freshworks: We want to make sure that you can delete any specialty selection, unless there is only one left.

dbayly-freshworks: We want to make sure that you can delete any specialty selection, unless there is only one left.

apps/web/src/components/submission/components/Credential.tsx
dbayly-freshworks: Good catch!  i will update this

dbayly-freshworks: Good catch!  i will update this

sunghwan-park-fw (approved):

PR #176, opened 2022-03-29
code-base review

sunghwan-park-fw:


packages/common/README.md
```diff
@@ -1 +1,16 @@
 # common
+
+This is a common project shared by api and web project. Shared components are mostly data models and validators.
```
dbayly-freshworks, line 3: Maybe use folder/directory instead of project

dbayly-freshworks (approved):
One small comment, otherwise looks great!

.github/workflows/deploy-test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     environment:
       name: test
     env:
-      CLOUDFRONT_ID: "${{ secrets.CLOUDFRONT_ID }}"
+      CLOUDFRONT_ID: '${{ secrets.CLOUDFRONT_ID }}'
```
vyasworks, line 18: Can you check why these are changing? Ideally then should not.

sunghwan-park-fw: `yarn format:write` changed it because `.prettierrc` enforces single quotes. Is it different if you run it on your env?

sunghwan-park-fw: #177 PR check also failed with the format check error caused by these double quotes.

PR #186, opened 2023-02-01
ehpr-260 add terraform code to deploy SES

sunghwan-park-fw:
- create AWS SES, DKIM
- grant access to SES to lambda


dbayly-freshworks (approved):
Looks good to me!

PR #188, opened 2023-02-02
ehpr-262 fix handling error of sending an mail with ses

sunghwan-park-fw:


apps/api/src/mail/mail.service.ts
```diff
@@ -124,7 +124,7 @@ export class MailService {
       Source: process.env.MAIL_FROM || 'EHPRDoNotReply@dev.ehpr.freshworks.club',
     };
     try {
-      return this.ses.sendEmail(params).promise();
+      return await this.ses.sendEmail(params).promise();
```
sunghwan-park-fw, line 127: Without `await`, this `try` cannot catch errors.

dbayly-freshworks (approved):

PR #189, opened 2023-02-03
ehpr-262 use fixed `from` address when using ches

sunghwan-park-fw:


apps/api/src/mail/mail.service.ts
```diff
@@ -32,7 +32,7 @@ export class MailService {
     if (!chesHost) return;
 
     const emailBody = {
-      from: mailOptions.from,
+      from: 'EHPRDoNotReply@gov.bc.ca',
```
sunghwan-park-fw, line 35: CHES seems to reject requests if the `from` is not `EHPRDoNotReply@gov.bc.ca`

dbayly-freshworks: I have mixed feelings about hard coding this. Could we instead just make sure that we're passing the correct email address where this function is called? 
Also shouldn't this be EHPRDoNotReply@ehpr.gov.bc.ca?

sunghwan-park-fw: It is just for CHES temporarily. One for SES is set by terraform.

dbayly-freshworks (approved):

PR #196, opened 2023-07-24
Completed work on EHPR-281 Add confirmation page for profile update

dbayly-freshworks:
Updated copy on landing page
Added confirmation page for after submission update.

### DIFF
--- a/apps/web/src/pages/submission/[step].tsx
+++ b/apps/web/src/pages/submission/[step].tsx
@@ -30,33 +30,27 @@ const Submission = () => {
           <section className='mb-7'>
             <p className='mb-4'>
               The <b>Emergency Health Provider Registry (EHPR)</b> is an online registry to support
-              the proactive deployment of health care providers to ensure BC’s health care system is
-              best prepared to respond to emergencies (e.g., wildfires, floods, pandemics). It is an
-              online registry of health care professionals and health authority staff who are able
-              and willing to be deployed or hired to support B.C.’s health system response. For more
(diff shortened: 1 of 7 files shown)


apps/web/src/pages/submission/[step].tsx
```diff
@@ -67,8 +61,10 @@ const Submission = () => {
             </p>
 
             <p className='font-bold'>
-              If you have any questions about our collection or use of personal information, please
-              email your inquiries to&nbsp;
+              <b>
```
Comment (coryatwater-FW) [inline apps/web/src/pages/submission/[step].tsx:64]: could probably remove this `<b>` tag

apps/web/src/pages/update-submission/confirmation.tsx
```diff
@@ -0,0 +1,38 @@
...
+const Confirmation: React.FC = () => {
+  const { push, query, isReady } = useRouter();
+  const id: string = query.id as string;
+
+  useEffect(() => {
+    // redirect to the home page if the id is not present
+    if (isReady && !id) {
+      push('/');
+    }
+  });
```
coryatwater-FW, line 14: how come the missing dependency as opposed to `id`?

dbayly-freshworks: I can clear this out. I copied this template from the confirmation page for the main form. An ID is not needed here.

coryatwater-FW: ah ok, was just curious if it was on purpose

coryatwater-FW (approved):

PR #205, opened 2023-09-07
EHPR-286 invite user

sunghwan-park-fw:
**Two ways for a new user to join EHPR**
1. Sign up by logging in and get approval
    1. If a new user logs in for the first time, an account is created with a `Pending` role.
    1. The administrator approves and assigns the user a `User` role.
1. Get an Invitation and log in
    1. The administrator sends an invitation email with the role assignment.
    2. A new account is created to be inactive.
    3. The account is activated by logging in with the same email address.

**Revocation**
The administrator can **revoke** a user by inactivating the account.

ehpr-invite

ehpr-invite-user


dbayly-freshworks (approved):
Looks good to me

PR #211, opened 2024-01-19
 EHPR-294 - add domain to api env variables

coryatwater-FW:
- added domain variable to environment list in `api.tf`


dbayly-freshworks (approved):
Looks good, feel free to merge after tests pass.

PR #214, opened 2024-02-09
EHPR-291 - add ability to send mass emails

coryatwater-FW:
- added ability to send mass emails
- added `Promise Pool` for concurrency measures, currently set to 5, can be changed.
- added confirmation modal *added TODO to add legal disclaimer
- added `mass_email_record` table to keep records of all mass email sends
- will create an entry whether successful or not, saving errors

### DIFF
--- a/apps/api/src/mass-email-record/mass-email-record.service.ts
+++ b/apps/api/src/mass-email-record/mass-email-record.service.ts
@@ -1,0 +1,63 @@
+import { Inject, Injectable, Logger } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { PromisePoolError } from '@supercharge/promise-pool';
+import { AppLogger } from '../common/logger.service';
+import { MassEmailRecordEntity } from './entity/mass-email-record.entity';
+import { GenericError } from 'src/common/generic-exception';
+
+@Injectable()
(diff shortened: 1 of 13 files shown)


apps/api/src/mass-email-record/mass-email-record.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    userId: string,
+    emailIds: string[],
+    errors: PromisePoolError<{ status: string; name: string; message: string; userId: string }>[],
+  ): Partial<MassEmailRecordEntity> {
+    let mappedErrors: GenericError[] = [];
+    // general exceptions and AWS errors have different formats
+    // check for general exceptions then look for AWS errors
+    if (errors) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      mappedErrors = errors.map(({ error, recipientId }: any) => {
```
dbayly-freshworks, line 42: Nit maybe use unknown instead of any here?

dbayly-freshworks (approved):
This looks ok to me. One small nit

apps/api/src/mass-email-record/mass-email-record.service.ts
coryatwater-FW: I'd still need to do some form of type checking with unknown,  I will take this and look into it

PR #221, opened 2024-02-21
EHPR-301 - add health authority table and assign ha id to users

coryatwater-FW:
- added migration for `health_authority` table
- add `ha_id` column and relation to `user` table
- update `ha_id` for existing users using their email domain
- prelim work for retrieving users HA for future filtering purposes

[EHPR-301](https://eydscanada.atlassian.net/jira/software/c/projects/EHPR/boards/260?selectedIssue=EHPR-301)


dbayly-freshworks (approved):
This looks ok to me

PR #224, opened 2024-02-22
add limiter library for aws ses sending rates

coryatwater-FW:
- added `limiter` library to reduce amount of requests per second to AWS
- set to 35 tokens/ requests per 1 second window
- AWS currently has a max of 40/s


dbayly-freshworks (approved):
This looks good to me

PR #231, opened 2024-03-07
make specialties list alphabetical

coryatwater-FW:
- make specialties list alphabetical again after adding `Paramedic` and `Firefighter`


dbayly-freshworks (approved):
Looks good to me

PR #234, opened 2024-03-14
EHPR-306 - add mass email feature flag

coryatwater-FW:
- add mass email feature flag to block access to mass email route and hide registrants table


dbayly-freshworks (approved):
Looks good to me 👍

PR #256, opened 2024-07-10
[EHPR] addressing CodeQL alerts

JGH-EY:
In the security tab CodeQl has new alerts with code

I addressed these as follows
- Enforce a format check for api call
- Made template replacement more secure and faster


dbayly-freshworks (approved):
Looks good to me

PR #262, opened 2024-11-25
EHPR - 339 Added filters for FNHA

dbayly-freshworks:
FNHA users should only be able to extract users who are willing to work anywhere.


jerry-ey (approved):
LGTM!
For the testing fail on github action, I think we can ignore. as my captcha PR did resolve that issue.

PR #265, opened 2024-11-29
EHPR-339 Added FNHA to full list of HAs

dbayly-freshworks:
Add FNHA to list of full HAs

Added a check to safely return when there is no submissions


jerry-ey (approved):
Looks great!

PR #272, opened 2025-02-13
Ehpr 335 enable filtered reporting 

dbayly-freshworks:
Changes 
1. Added Stream multi-select dropdown on extract page
2. Added Specialty multi-select dropdown on extract page
3. Added Subspecialty multi-select dropdown on extract page
4. Added location multi select drop down 
5. Location multi select is only shown if the user is MoH
6. Updated extract to apply filters 

Note - Parsing submission json down to levels to get subspecialties proved difficult. I put the filtering logic outside of the query.


jerry-ey (approved):
Looks good!

PR #276, opened 2025-03-05
EHPR-320 logging mass email with subject and message id

sunghwan-park-fw:


dbayly-freshworks (approved):
This looks good. Approved!

PR #278, opened 2025-03-06
EHPR-344 view/download mass email history

sunghwan-park-fw:
- Add mass email history tab to admin page
- Add modal to view full recipients/errors
- Download full history as csv
- EHPR-336 update stream/specialty

image
image
image
[ehpr-mass-email-history-2025-03-06.csv](https://github.com/user-attachments/files/19114705/ehpr-mass-email-history-2025-03-06.csv)


apps/api/src/registrant/registrant.controller.ts
```diff
@@ -51,12 +57,31 @@ export class RegistrantController {
     return registrants;
   }
 
+  @Roles(Role.Admin, Role.User)
+  @UseGuards(AuthGuard, RoleGuard)
+  @UseInterceptors(ClassSerializerInterceptor)
+  @Get('/mass-email')
```
dbayly-freshworks, line 63: Maybe change this to /mass-email-history for clarity in debugging?

dbayly-freshworks (approved):
Overall this looks good. Two small comments. 

I had some concerns about merging this work in with the stream updates. With the new filters, applicants who selected a removed/renamed stream might be unfairly excluded from searches. I checked the registry, and it looks like only 36 users are IENs, most belonging to other streams. Considering the registry has ~7k users, I think the impact will be minimal.

apps/api/src/registrant/registrant.controller.ts
sunghwan-park-fw: updated.

PR #292, opened 2025-07-25
JSU79: Update global JS versions for repo-support packages

npham49:
Description of changes:
- Adding mass email feature to env sample file
- Update to Prettier (3.6.2 + support package pretty-quick), Typescript (5.8.3), Husky (9.1.7), and corresponding code changes
- Adding option to use a local SES instance for local dev to prevent incurring extra costs for client
- Update null catching on Mass Email History table, this page was running into an error every time developers try to load without sending a mass email


gperkinscgi (approved):
Reviewed changes, no concerns

PR #311, opened 2026-04-28
set width and height attributes for the moh logo directly

chloe-yuu:
set width and height attributes for the moh logo directly


lucianod-bcgov (approved):
Looks good. Issue is on Test so did not test locally.

